Issue #11, opened 2019-12-11
HTTP Check


moshloop:
**Input**:

```yaml
# If DNS, test connectivity for each returned record
endpoints: 
   - https://google.com
   - http://google.com:443
# max response time
thresholdMillis: 300
responseCodes: [201,200,301]
responseContent: "OK"
maxSSLExpiry: 60d

```

**Metrics** (per endpoint)
* response_time
* response_code
* ssl_certificate_expiry

PR #18, opened 2019-12-23
#11 implemented config reader and http checks

K-DOT:


test.yaml
```diff
@@ -0,0 +1,13 @@
+http:
```
moshloop: Can we add test file similar to https://github.com/flanksource/konfigadm/blob/master/test/setup_test.go#L161 and then move this files into a test or fixtures folder?


We probably also need to split into 2 sets of fixtures, 1 for passing checks and 1 for failing.

pkg/checker.go
```diff
@@ -0,0 +1,155 @@
...
+
+func dnsLookup(endpoint string) []URL {
+	var result []URL
+	parsedURL, err := url.Parse(endpoint)
+	if err != nil {
+		log.Fatal(err)
+	}
+	ips, err := net.LookupIP(parsedURL.Hostname())
+	if err != nil {
+		log.Fatal(err)
```
moshloop: return the error e.g. (`func dnsLookup(endpoint string) ([]URL, error)`, rather than exiting (`log.Fatal`)

pkg/config.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+
+	"gopkg.in/yaml.v2"
+)
+
+// ReadConfig : Read config and call CheckConfig
+func ReadConfig(configfile string) Config {
```
moshloop: This should probably be split into a ParseConfig and RunChecks or similar method

pkg/checker.go
```diff
@@ -0,0 +1,155 @@
...
+					}
+					result = append(result, checkResult)
+				}
+			}
+		}
+	}
+	return result
+}
+
+func checkHTTP(urlObj URL) (interface{}, error) {
```
moshloop: return `HTTPCheckResult` and not interface{}

pkg/checker.go
```diff
@@ -0,0 +1,155 @@
...
+	ip     string
+	port   int
+	host   string
+	scheme string
+	path   string
+}
+
+// CheckConfig : Check every record of DNS name agaist config information
+// Returns check result and metrics
+func CheckConfig(args map[string]interface{}) []*CheckResult {
```
moshloop: this method should probably be moved to a`http` package and renamed `Check(check HttpCheck) []*CheckResult`

pkg/config.go
```diff
@@ -0,0 +1,31 @@
...
+	data, err := ioutil.ReadFile(configfile)
+	if err != nil {
+		log.Printf("yamlFile.Get err   #%v ", err)
+	}
+	yamlerr := yaml.Unmarshal([]byte(data), &config)
+	if yamlerr != nil {
+		log.Fatalf("error: %v", yamlerr)
+	}
+	for _, conf := range config.HTTP {
+		m := map[string]interface{}{
```
moshloop: this block should be replaced with a `http.Check(conf)`

pkg/config.go
```diff
@@ -0,0 +1,31 @@
...
+			"http": conf,
+		}
+		for _, i := range CheckConfig(m) {
```
moshloop: i should only be used for index variables, i.e. rename to result.

Issue #14, opened 2019-12-11
Prometheus Export


moshloop:
`/metrics`

Export a snapshot of metrics created by all configured checks

PR #19, opened 2020-01-08
Add prometheus metrics

teodor-pripoae:
Solves #14


http/check.go
```diff
@@ -3,16 +3,53 @@ package http
...
 	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/flanksource/canary-checker/pkg"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	opsProcessed = prometheus.NewCounter(prometheus.CounterOpts{
```
moshloop: Can we decouple the check from prometheus via the api.Metric ? 

Ops can be inferred, and if we add Duration to  https://github.com/flanksource/canary-checker/blob/master/pkg/api.go#L28 then it leaves just status_code and dnsFailed as custom metrics

PR #28, opened 2020-03-04
Add s3 bucket scan

teodor-pripoae:


pkg/api.go
```diff
@@ -100,6 +101,21 @@ type S3Check struct {
 	ObjectPath string   `yaml:"objectPath"`
 }
 
+type S3BucketCheck struct {
+	Bucket    string `yaml:"bucket"`
```
moshloop: Can we make this support env variable like: https://github.com/flanksource/canary-checker/blob/master/pkg/config.go#L30

pkg/api.go
```diff
@@ -100,6 +101,21 @@ type S3Check struct {
...
+	AccessKey string `yaml:"accessKey"`
+	SecretKey string `yaml:"secretKey"`
+	Region    string `yaml:"region"`
+	Endpoint  string `yaml:"endpoint"`
+	// glob path to restrict matches to a subet
+	ObjectPath string `yaml:"objectPath"`
+	ReadWrite  bool   `yaml:"readWrite"`
+	// maximum allowed age of matched objects
+	MaxAge int64 `yaml:"maxAge"`
```
moshloop: Can we add units to these descriptions?

PR #29, opened 2020-03-04
Add LDAP check

teodor-pripoae:


checks/ldap.go
```diff
@@ -0,0 +1,105 @@
...
+	"fmt"
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/flanksource/canary-checker/pkg"
+	ldap "github.com/go-ldap/ldap/v3"
+)
+
+var (
+	ldapLookupHistogram = prometheus.NewHistogramVec(
```
moshloop: Can't we use the default Duration metric for this?

Issue #6, opened 2019-12-11
Postresql SQL execute


moshloop:
**Input**
```yaml
connection: user=pqgotest dbname=pqgotest sslmode=verify-full
query:  SELECT 1
results: 1
```

**Output**: Pass/Fail/Invalid

**Metrics**:

* response_time
* Record count (min,max,avg)

philipstaffordwood:
working on [this branch and fork](https://github.com/flanksource/canary-checker/compare/master...philipstaffordwood:add-postgres-check?expand=1)

PR #31, opened 2020-03-05
Add postgres check

philipstaffordwood:
This PR adds basic postgres connectivity testing as suggested by https://github.com/flanksource/canary-checker/issues/6

For now it allows arbitrary SQL that returns **single row single value results**.

> The design intent is understood to be more that of connectivity testing (as opposed to arbitrary SQL query execution and result interpretation).
> (Happy to extend on the base functionality introduced by this PR in a subsequent PR to introduce more flexible result handling)


checks/postgres.go
```diff
@@ -0,0 +1,125 @@
...
+	"time"
+
+	"github.com/flanksource/canary-checker/pkg"
+	_ "github.com/lib/pq"
+	"github.com/prometheus/client_golang/prometheus"
+	log "github.com/sirupsen/logrus"
+)
+
+var (
+	postgresFailed = prometheus.NewCounter(prometheus.CounterOpts{
```
moshloop: Why are these 2 metrics required? Do they not duplicate the standard metrics which provide a duration and failed metric by default ?

go.mod
```diff
@@ -3,7 +3,9 @@ module github.com/flanksource/canary-checker
 go 1.12
 
 require (
+	github.com/DATA-DOG/go-sqlmock v1.4.1
 	github.com/Microsoft/go-winio v0.4.14 // indirect
+	github.com/apex/log v1.1.2
```
moshloop: this module should be unnecessary

philipstaffordwood: yes, fixed with 58d62537a5403a106a72aa16c9a320079c946b0b

checks/postgres.go
philipstaffordwood: :man_facepalming: yes!

philipstaffordwood: removed with `canary_check_postgres_query_time` metric with 220deb992a75adf5a887bfcc31668cc1417ab26b

philipstaffordwood: removed `canary_check_postgres_connectivity_failed` with 76e9128b91c5677bda638fe6717f90b41f290ac8

philipstaffordwood:
@moshloop 

The original ask is for:
>**Metrics:**
>
>response_time
>Record count (min,max,avg)

The default `canary_check_duration` covers `response_time`, right?

With the scaffolding in place should I extend the checking to produce another meaningful metric?

I could extend the query result handling to retrieve more generic multi-value, multi-row results and then add the **Record count (min,max,avg)** metrics.

Are we intending to check just basic connectivity, or system/deployment health via some kind of heartbeat like query?

moshloop:
The purpose of this would be to check for both basic connectivity and more complex health via a heartbeat-like query.

PR #61, opened 2020-06-24
Set pod priority for pod canaries

teodor-pripoae:


moshloop:
Can we add the canary priority class to the k8 deployment manifest?

teodor-pripoae:
Added in https://github.com/flanksource/karina/pull/398

PR #92, opened 2020-10-08
documentation updates.  Added prerequistes and introduction

vallard:
I want to see if I'm on the right track with this.  I started trying to move the documentation into the official docs: https://canary-checker.docs.flanksource.com and pointing the README to this page.  I also started explaining the need for synthetic monitoring and added more information on how to get the prerequisites installed.  
In addition I created a few checks and uploaded an image of my checks as part of the documentation.  
I'm still not clear on the installation guide as whether the binaries should be downloaded to use as stand alone outside of Kubernetes or as a client with the canary checker that is running in my kubernetes cluster.  Thanks for any guidance!


README.md
```diff
@@ -9,12 +9,14 @@
 <a href="https://canary-checker.docs.flanksource.com"> <img src="https://img.shields.io/badge/☰-Docs-lightgrey.svg"/> </a>
 </p>
 
+## Documentation
```
moshloop, line 12: I think use `Kubernetes native` rather than native to Kubernetes.

README.md
```diff
@@ -9,12 +9,14 @@
...
+
+Canary Checker is a multi-tenant synthetic monitoring system native to Kubernetes.  To learn more, please see the [official documentation](https://canary-checker.docs.flanksource.com).
+
 
 ---
 <!--ts-->
-  * [Features](#features)
```
moshloop, line 15: I think the Features/Comparisons should be included or duplicated here as the README is often the landing page and we need to "sell" users on why the would even want to go deeper into the docs.

docs/prereqs.md
```diff
@@ -0,0 +1,83 @@
...
+To ensure you are getting metrics run: 
+
+```
+kubectl get --raw /metrics
+```
+
+
+## Prometheus Operator
+
+The Prometheus Operator is required for Canary Checker.
```
moshloop, line 22: So this isn't strictly true - There are service monitors that are bundled in the deployment file, but other than that there is no dependency.  I also don't think we need to include the installation instructions

vallard: When I installed I got an error stating service monitor crds were not found that were resolved by installing Prometheus. I will say that it is optional.  I know I got value out of installing the operator installation just because I spent about an hour trying to find the easiest way to install it.  Everywhere I saw suggested helm and I didn't want to use helm.

Issue #90, opened 2020-10-06
Add header for namespace, and remove namespace prefix in name


moshloop:

philipstaffordwood:
We discussed not having a **Namespace** Column and rather relying on grouping
![image](https://user-images.githubusercontent.com/11496482/98297167-ace60300-1fbc-11eb-8e46-c5e6876de02b.png)

But adding a "indentation" under the namespace for the grouped checks seems to look cleaner, and that layout is easier using an extra column 
![image](https://user-images.githubusercontent.com/11496482/98297491-2b42a500-1fbd-11eb-8e23-68f87de1bf1c.png)

(and if we have it we may as well label it for clarity)
![image](https://user-images.githubusercontent.com/11496482/98296697-f41fc400-1fbb-11eb-9d4d-90b9de8a0e62.png)

moshloop:
Looks good - I think we should switch the tag from the name to namespace, and also reduce the width of the column to maybe 50 - 100px , The type column width can also be reduced

philipstaffordwood:
OK:
Tag on namespace, not name,
Columns sized to fit contents,
Border cleanups, and
A bit of spacing between each namespace section

![image](https://user-images.githubusercontent.com/11496482/98451478-77711f00-214e-11eb-87f4-af820272fdd5.png)

Issue #105, opened 2020-11-05
Duplicate Vue keys for http checks with same endpoint with different responseCodes


philipstaffordwood:
Duplicate `key`s are produced for `checkSets` for **http** checks in the GUI if same `endpoint` with different `responseCodes` is tested causing Vue updating issues.

e.g.

```YAML
apiVersion: canaries.flanksource.com/v1
kind: Canary
metadata:
  name: http-miltple
spec:
  interval: 30
  http:
    - endpoint: https://httpstat.us/500
      thresholdMillis: 3000
      responseCodes: [500]
      responseContent: ""
      maxSSLExpiry: 7
    - endpoint: https://httpstat.us/500
      thresholdMillis: 3000
      responseCodes: [302]
      responseContent: ""
      maxSSLExpiry: 7
```

PR #106, opened 2020-11-07
Add header for namespace, and remove namespace prefix in name

philipstaffordwood:
Closes https://github.com/flanksource/canary-checker/issues/90

image


pkg/api.go
```diff
@@ -18,7 +18,7 @@ type Endpoint struct {
 type JSONTime time.Time
 
 func (t JSONTime) MarshalJSON() ([]byte, error) {
-	stamp := fmt.Sprintf("\"%s\"", time.Time(t).Format("2006-01-02 15:04:05"))
+	stamp := fmt.Sprintf("\"%s\"", time.Time(t).Format("2006-01-02 15:04:05.001"))
```
philipstaffordwood, line 21: Upped the time resolution to prevent duplicate keys in `checkSets`.
Also logged https://github.com/flanksource/canary-checker/issues/105 for **http** checks with same **endpoint** with different **responseCodes** that cause [duplicate keys](https://github.com/philipstaffordwood/canary-checker/blob/5708ccdbd5ec74859c522830b83bcd12c344f38e/statuspage/src/store/index.js#L29)

PR #113, opened 2021-02-03
additional debug output

BrendanGalloway:
Don't merge, just testing


BrendanGalloway:
It looks like I've tracked down some of the issues that were causing failures in the github actions:

the KUBECONFIG was not being passed into the sudo environment running the test, meaning the k9s clients were not created correctly
Some tests returned nil when encountering a lock, but higher up in the stack no logic was present to cater for nil CheckResults being returned

In this process, the kommons client k8s client creation system was incorporated, and all the k8s api calls were updated to v19.  I can probably split this part out into a separate PR if it's strictly necessary since it ended up not actually fixing any of the testing problems, but would need to be done at some point in any case...

PR #117, opened 2021-02-08
Apply full gomplate parsing to canary inputs

BrendanGalloway:


pkg/config.go
```diff
@@ -20,11 +22,16 @@ func ParseConfig(configfile string) v1.CanarySpec {
 	if err != nil {
 		logger.Infof("yamlFile.Get err   #%v ", err)
 	}
-	yamlerr := yaml.Unmarshal([]byte(data), &config)
+	envVar := GetEnvMap()
+	parsed, err := ApplyGomplate(string(data), envVar)
```
moshloop, line 26: So this templates the entire spec, which we don't want - we only want to template each value in StructTemplater. 

We also need to cater for the controller here: https://github.com/flanksource/canary-checker/blob/master/pkg/controllers/canary_controller.go#L245

BrendanGalloway:
Abandoned in favour of #118

BrendanGalloway:
Abandoned in favour of #118

Issue #131, opened 2021-03-06
HTTP Check Enhancements


moshloop:
Add 

* username/password from secret for basic auth
* headers / inline or from secret/configmap
* method
* body
* responseJSONPath expression instead of responseContent

PR #135, opened 2021-03-11
Httpcheck improvements

BrendanGalloway:
#131 

@moshloop Please specifically check if you are happy with how I am parsing from env var.


checks/http.go
```diff
@@ -43,7 +48,33 @@ func init() {
...
-type HttpChecker struct{}
+type HttpChecker struct {
+	k8s           *kommons.Client
+	specNamespace string
+}
+
+func NewHttpChecker() *HttpChecker {
+	nc := &HttpChecker{}
+
+	kommons, err := pkg.NewKommonsClient()
```
moshloop, line 59: We don't want todo this at startup, only when an actual http check is run, so this probably needs to be on HttpCheck rather than HttpChecker

checks/http.go
```diff
@@ -67,6 +97,8 @@ func (c *HttpChecker) Check(extConfig external.Check) *pkg.CheckResult {
 	check := extConfig.(v1.HTTPCheck)
 	endpoint := check.Endpoint
 	namespace := check.Namespace
+	var username string
+	var password string
```
moshloop, line 101:
```suggestion
```

pkg/config.go
```diff
@@ -24,7 +24,8 @@ func ParseConfig(configfile string) v1.CanarySpec {
 	if yamlerr != nil {
 		logger.Fatalf("error: %v", yamlerr)
 	}
-	return ApplyTemplates(config)
+	return config
```
moshloop, line 27: is this intentional ?

checks/http.go
```diff
@@ -67,6 +80,12 @@ func (c *HttpChecker) Check(extConfig external.Check) *pkg.CheckResult {
 	check := extConfig.(v1.HTTPCheck)
 	endpoint := check.Endpoint
 	namespace := check.Namespace
+	specNamespace := check.GetNamespace()
+
+	kommons := c.GetClient()
+	if kommons == nil {
+		return Failf(check, "Kommons client not set for HttpChecker instance")
```
moshloop, line 87: This should only fail when actually trying to use the client

PR #154, opened 2021-04-08
feat: Add popup on a canary with its details

### LINKED ISSUE #136: Add severity and owner/email
Add on dashboard icons on the severity and popup with owner/email details.
+ Filter on severity / owner
+ add severity / email details into the prometheus alerts

```
kind: Canary
metadata:
  name: http-pass
spec:
  interval: 30
  description: "When this canary fails it means that either http.us is down (unknlikely) or there is an issue with upstream i internet connection"
  owner: networks
  severity: critical
  http:
    - endpoint: https://httpstat.us/200
      thresholdMillis: 3000
      responseCodes: [201, 200, 301]
      responseContent: ""
      maxSSLExpiry: 7
```

Kaitou786:
closes #136

Added a pop-up for the canary. When someone clicks on any canary a popup will open with the canary details
The popup looks like:
Screenshot 2021-04-08 at 10 36 32 AM

### DIFF
--- a/api/v1/canary_types.go
+++ b/api/v1/canary_types.go
@@ -43,6 +43,8 @@ type CanarySpec struct {
 	Namespace      []NamespaceCheck      `yaml:"namespace,omitempty" json:"namespace,omitempty"`
 	Redis          []RedisCheck          `yaml:"redis,omitempty" json:"redis,omitempty"`
 	Interval       uint64                `yaml:"interval,omitempty" json:"interval,omitempty"`
+	Severity       string                `yaml:"severity,omitempty" json:"severity,omitempty"`
+	Owner          string                `yaml:"owner,omitempty" json:"owner,omitempty"`
 }
 
 func (spec CanarySpec) GetAllChecks() []external.Check {
--- a/statuspage/src/App.vue
+++ b/statuspage/src/App.vue
@@ -21,7 +21,7 @@
                 </tr>
 
                 <template v-for="(check) in byNamespace.items" >
-                      <table-row :interval="check.interval"
+                      <canary-standard-server-row :interval="check.interval"
                                  :owner="check.owner"
                                  :severity="check.severity"
                                  :check-type="check.type"
@@ -54,12 +54,12 @@ import Vuex from "vuex";
 import store from "./store";
 import AutoUpdateSettings from "./components/AutoUpdateSettings.vue";
 import ErrorPanel from "./components/ErrorPanel.vue";
-import TableRow from "@/components/TableRow";
+import CanaryStandardServerRow from "./components/CanaryStandardServerRow.vue";
 
 export default {
   name: "App",
   components: {
-    TableRow,
+    CanaryStandardServerRow,
     AutoUpdateSettings,
     ErrorPanel,
   },
--- a/statuspage/src/App.vue
+++ b/statuspage/src/App.vue
@@ -22,8 +22,9 @@
 
                 <template v-for="(check) in byNamespace.items" >
                     <tr :key="checkKey(check)" >
-                        <td v-b-modal="'modal-canary'+check.name+check.namespace">
-                                <img :src="'images/' + check.type + '.svg'" :title="check.type " height="20px" >  {{ shortHand(checkName(check),60) }}
+<!--                      Component modal doc: : https://bootstrap-vue.org/docs/components/modal-->
+                      <td v-b-modal="`modal-canary${check.name}${check.namespace}`">
+                                <img :src="`images/${check.type}.svg`" :title="check.type " v-bind:style="{ height: '1.25rem' }" alt="" >  {{ shortHand(checkName(check),60) }}
                               <canary-modal :interval="check.interval"
                                             :owner="check.owner"
                                             :severity="check.severity"
@@ -115,8 +116,7 @@ export default {
 
     checkKey() {
       return (check) => {
-        let id = check.key + check.id + check.description + check.name + check.namespace + check.endpoint + check.serverURL
-        return id
+        return `${check.key}${check.id}${check.description}${check.name}${check.namespace}${check.endpoint}${check.serverURL}`
       }
     },
 
--- a/statuspage/src/App.vue
+++ b/statuspage/src/App.vue
@@ -21,29 +21,17 @@
                 </tr>
 
                 <template v-for="(check) in byNamespace.items" >
-                    <tr :key="checkKey(check)" >
-<!--                      Component modal doc: : https://bootstrap-vue.org/docs/components/modal-->
-                      <td v-b-modal="`modal-canary${check.name}${check.namespace}`">
-                                <img :src="`images/${check.type}.svg`" :title="check.type " v-bind:style="{ height: '1.25rem' }" alt="" >  {{ shortHand(checkName(check),60) }}
-                              <canary-modal :interval="check.interval"
-                                            :owner="check.owner"
-                                            :severity="check.severity"
-                                            :check-type="check.type"
-                                            :description="check.description"
-                                            :name="check.name"
-                                            :namespace="check.namespace"
-                              />
-                        </td>
-
-                        <td v-for="server in orderedServers" :key="server.value" class="align-top border-right border-left">
-                            <div>
-                              <status-strip  :checks="check.items" :server="server.value"
-                                          color="#28a745" error-color="#dc3545"
-                                           :bar-width="20" :bar-spacing="5" :barMaxHeight="20"
-                                           :zoominess="0.85"/>
-                            </div>
-                        </td>
-                    </tr>
+                      <table-row :interval="check.interval"
+                                 :owner="check.owner"
+                                 :severity="check.severity"
+                                 :check-type="check.type"
+                                 :description="check.description"
+                                 :name="check.name"
+                                 :namespace="check.namespace"
+                                 :short-description="shortHand(checkName(check),60)"
+                                 :items="check.items"
+                                 :key="checkKey(check)"
+                                 />
                 </template>
 
             </template>
@@ -66,16 +54,14 @@ import Vuex from "vuex";
 import store from "./store";
 import AutoUpdateSettings from "./components/AutoUpdateSettings.vue";
 import ErrorPanel from "./components/ErrorPanel.vue";
-import StatusStrip from "./components/StatusStrip.vue";
-import CanaryModal from "@/components/CanaryModal";
+import TableRow from "@/components/TableRow";
 
 export default {
   name: "App",
   components: {
-    CanaryModal,
+    TableRow,
     AutoUpdateSettings,
     ErrorPanel,
-    StatusStrip,
   },
   store: store,
   created() {
--- a/statuspage/src/App.vue
+++ b/statuspage/src/App.vue
@@ -22,8 +22,16 @@
 
                 <template v-for="(check) in byNamespace.items" >
                     <tr :key="checkKey(check)" >
-                        <td>
-                                <img :src="'images/' + check.type + '.svg'" :title="check.type " height="20px">  {{ shortHand(checkName(check),60) }}
+                        <td v-b-modal="'modal-canary'+check.name+check.namespace">
+                                <img :src="'images/' + check.type + '.svg'" :title="check.type " height="20px" >  {{ shortHand(checkName(check),60) }}
+                              <canary-modal :interval="check.interval"
+                                            :owner="check.owner"
+                                            :severity="check.severity"
+                                            :check-type="check.type"
+                                            :description="check.description"
+                                            :name="check.name"
+                                            :namespace="check.namespace"
+                              />
                         </td>
 
                         <td v-for="server in orderedServers" :key="server.value" class="align-top border-right border-left">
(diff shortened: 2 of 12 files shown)


statuspage/src/components/CanaryModal.vue
```diff
@@ -0,0 +1,60 @@
+<!-- This component encapsulates a modal that is displayed -->
+<!-- when clicking on canary                               -->
+<template>
+  <b-modal :id="'modal-canary'+name+namespace" ok-only>
+    <template #modal-title>
```
flanksaul, line 5: We'll need a way of maintaining accessibility if we're going to use this approach. We need to ensure that the `aria-labelledby` attribute points at an appropriate node: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute

See bootstrap notes: https://bootstrap-vue.org/docs/components/modal#accessibility

statuspage/src/components/CanaryModal.vue
```diff
@@ -0,0 +1,60 @@
...
+      <img :src="'images/' + checkType + '.svg'" height="30px"> {{name}}
+    </template>
+    <template v-slot:default>
```
flanksaul, line 8: I don't believe we need this `<template>` node.

flanksaul (changes requested):
More comments to follow.

statuspage/src/components/CanaryModal.vue
```diff
@@ -0,0 +1,60 @@
...
+<script>
+
+
+
+export default {
+  name: "CanaryModal",
+  props: {
+    interval: {
+      type: String,
+      required: true
```
flanksaul, line 28: Just to note here: these do not throw errors, they just log to console.

statuspage/src/store/index.js
```diff
@@ -135,7 +135,11 @@ export default new Vuex.Store({
                     group[check.namespace][groupBy] = {
                         type: check.type,
                         namespace: check.namespace,
-                        name: check.description ? check.description : check.name,
```
flanksaul, line 138: Without e2e tests, would like a view on whether we want to maintain this fallback from whomever knows best.

statuspage/src/store/index.js
```diff
@@ -135,7 +135,11 @@ export default new Vuex.Store({
                     group[check.namespace][groupBy] = {
                         type: check.type,
                         namespace: check.namespace,
```
flanksaul, line 137: For syntax see:
- Nullish coalescing operator (??): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator
- Object Destructuring: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

We could approach this section in a couple of ways:

In the first, we're setting defaults with the nullish coalescing operator (if we want defaults).

```
   const { type, namespace, name, interval } = check;

      group[namespace][groupBy] = {
          type: type ?? "",
          namespace: namespace ?? "",
          name: name ?? "",
          interval: interval ?? 0,
          items: []
      }
```

But if we're confident of the values, we might do the following:

```
   const { type, namespace, name, interval } = check;
   group[namespace][groupBy] = { type, namespace, name, interval, items: [] }    
```

If we destructure at the top of the loop, we can use the variables all the way down, also, so:

`        
if (group[check.namespace] == null) {
            group[check.namespace] = {}
 }
`
                
Could be, for example:

`        
 group[namespace] = group[namespace] ?? {}
`

And:


```
if (group[check.namespace][groupBy] == null) {
      group[check.namespace][groupBy] = {
          type: check.type,
          namespace: check.namespace,
          name: check.name,
          interval: check.interval,
          severity: check.severity,
          owner:    check.owner,
          description: check.description,
          items: []
      }
}
```


Could be, at its most terse:

`
group[namespace][groupBy] = group[namespace][groupBy] ?? { type, namespace, name, interval, items: [] }
`

flanksaul (changes requested):
More coming still

statuspage/src/components/CanaryModal.vue
```diff
@@ -0,0 +1,60 @@
+<!-- This component encapsulates a modal that is displayed -->
+<!-- when clicking on canary                               -->
+<template>
+  <b-modal :id="'modal-canary'+name+namespace" ok-only>
+    <template #modal-title>
+      <img :src="'images/' + checkType + '.svg'" height="30px"> {{name}}
```
flanksaul, line 6: Should use template literals. Then, `rem` values over `px` values are preferred as `rem` can be scaled against base font size. See: https://nekocalc.com/px-to-rem-converter

Change the base font size to see what I mean.

At some stage, if `rem`s are consistently used, we can use the base font size to scale everything, meaning we'd need fewer media queries.

api/v1/canary_types.go
```diff
@@ -43,6 +43,8 @@ type CanarySpec struct {
 	Namespace      []NamespaceCheck      `yaml:"namespace,omitempty" json:"namespace,omitempty"`
 	Redis          []RedisCheck          `yaml:"redis,omitempty" json:"redis,omitempty"`
 	Interval       uint64                `yaml:"interval,omitempty" json:"interval,omitempty"`
+	Severity       string                `yaml:"severity,omitempty" json:"severity,omitempty"`
+	Owner          string                `yaml:"owner,omitempty" json:"owner,omitempty"`
```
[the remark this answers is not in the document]
flanksaul: I understand now that leaving out `omitempty` sets a field to required in the `crd.yaml` generation.

statuspage/src/components/CanaryModal.vue
```diff
@@ -0,0 +1,60 @@
...
+    </template>
+    <template v-slot:default>
+      <div class="name"><b>Name:</b> {{name}}</div>
```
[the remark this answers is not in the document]
Kaitou786: Change to `no-style` list

statuspage/src/components/CanaryModal.vue
```diff
@@ -0,0 +1,60 @@
...
+      type: String,
+      required: true,
+    },
+  },
+}
+</script>
+
+<style scoped>
+
+</style>
```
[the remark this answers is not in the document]
Kaitou786: Now using :)

statuspage/src/App.vue
```diff
@@ -22,8 +22,16 @@
 
                 <template v-for="(check) in byNamespace.items" >
                     <tr :key="checkKey(check)" >
-                        <td>
-                                <img :src="'images/' + check.type + '.svg'" :title="check.type " height="20px">  {{ shortHand(checkName(check),60) }}
+                        <td v-b-modal="'modal-canary'+check.name+check.namespace">
```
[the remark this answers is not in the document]
Kaitou786: Changed to template literal

statuspage/src/App.vue
```diff
@@ -22,8 +22,16 @@
...
+                                <img :src="'images/' + check.type + '.svg'" :title="check.type " height="20px" >  {{ shortHand(checkName(check),60) }}
```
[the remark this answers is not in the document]
Kaitou786: Defined

statuspage/src/components/CanaryModal.vue
```diff
@@ -0,0 +1,60 @@
...
+    <template #modal-title>
+      <img :src="'images/' + checkType + '.svg'" height="30px"> {{name}}
+    </template>
+    <template v-slot:default>
+      <div class="name"><b>Name:</b> {{name}}</div>
+      <div class="namespace"><b>Namespace:</b> {{namespace}}</div>
+      <div class="description"><b>Description:</b> {{description}}</div>
+      <div class="interval"><b>Interval:</b> {{interval}} Seconds</div>
+      <div class="owner" v-if="owner.length !== 0"><b>Owner:</b> {{owner}}</div>
+      <div class="severity" v-if="severity.length !== 0"><b>Severity:</b> {{severity}}</div>
```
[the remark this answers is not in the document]
Kaitou786: Changed to maintain structure integrity so if no severity or owner is provided we put a `-` over there see below image:
Screenshot 2021-04-08 at 8 23 00 PM

statuspage/src/components/CanaryModal.vue
Kaitou786: removed

statuspage/src/store/index.js
Kaitou786: For now changed to use `Nullish coalescing operator` maybe the possibility to use `reduce` can be explored when we will refactor this whole component

statuspage/src/components/CanaryModal.vue
Kaitou786: Changed :)

statuspage/src/components/CanaryModal.vue
moshloop: Can we add accessibility handling into the style guide?  It might be something we explicitly defer until we have a hard requirement and/or VC funding

flanksaul: @moshloop It's not easy to bolt on after the fact so would be ideal if it could be part of habitual best practice. 

What do you mean add it into the style guide in this case?

Either way, I did some research last night, and in this case if we make the `alt` tag on the image `alt=""` then screen reader should skip it when reading, so instead of reading `Heading Level 5 undefined image <title>` it would just read `Heading Level 5 <title>`, which was the behaviour from the unmodified component.

statuspage/src/App.vue
```diff
@@ -57,14 +54,14 @@ import Vuex from "vuex";
 import store from "./store";
 import AutoUpdateSettings from "./components/AutoUpdateSettings.vue";
 import ErrorPanel from "./components/ErrorPanel.vue";
-import StatusStrip from "./components/StatusStrip.vue";
+import TableRow from "@/components/TableRow";
 
 export default {
   name: "App",
   components: {
+    TableRow,
```
flanksaul, line 62: Maybe `CanaryStandardServerRow`? `TableRow` feels like it covers too much territory. Otherwise, nice work on this, you're picking up Vue quickly. :)

statuspage/src/App.vue
```diff
@@ -105,8 +102,7 @@ export default {
 
     checkKey() {
       return (check) => {
-        let id = check.key + check.id + check.description + check.name + check.namespace + check.endpoint + check.ServerURL
-        return id
+        return `${check.key}${check.id}${check.description}${check.name}${check.namespace}${check.endpoint}${check.serverURL}`
```
flanksaul, line 105: Appreciate the tidy.

statuspage/src/components/CanaryModal.vue
```diff
@@ -0,0 +1,73 @@
+<!-- This component encapsulates a modal that is displayed -->
+<!-- when clicking on canary                               -->
+<template>
+  <!--  Component modal doc: : https://bootstrap-vue.org/docs/components/modal-->
```
flanksaul, line 4: I would add a comment at the `v-b-modal` declaration too, even if more terse, as that would be the first encounter. 

This is good though, so leave it in. Just maybe add another pointer.

statuspage/src/components/CanaryModal.vue
```diff
@@ -0,0 +1,73 @@
...
+      required: true,
+    },
+    checkType: {
+      type: String,
+      required: true,
+    },
+  },
+  methods: {
+    getProp(prop){
+      if (prop.length === 0) {
```
flanksaul, line 58:
```suggestion
      if (typeof prop === 'string' && prop.length === 0) {
```
Ints and bools won't have the length property.

See the typeof operator: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof

statuspage/src/components/CanaryModal.vue
```diff
@@ -0,0 +1,73 @@
...
+  <b-modal :id="`modal-canary${name}${namespace}`" ok-only>
+    <template #modal-title>
+      <img :src="`images/${checkType}.svg`" v-bind:style="{ height: '1.875rem' }" :alt="`${checkType} logo`"> {{name}}
+    </template>
+    <ul class="list-unstyled">
+      <li><b>Name:</b> {{name}}</li>
+      <li><b>Namespace:</b> {{namespace}}</li>
+      <li><b>Description:</b> {{description}}</li>
+      <li><b>Interval:</b> {{interval}} Seconds</li>
+      <li><b>Owner:</b> {{getProp(owner)}}</li>
```
flanksaul, line 16: Maybe add the test for all of them?

statuspage/src/components/TableRow.vue
```diff
@@ -0,0 +1,82 @@
+<!-- This component encapsulates the table rows that is displayed -->
+<!-- when clicking on canary                               -->
+<template>
```
flanksaul, line 3: Appreciate this, thanks. Nice work on the Vue.

statuspage/src/App.vue
Kaitou786: Thanks and changed :)

statuspage/src/App.vue
Kaitou786: Thanks :)

statuspage/src/components/CanaryModal.vue
Kaitou786: Added a comment over there as well :)

statuspage/src/components/CanaryModal.vue
Kaitou786: Done

statuspage/src/components/TableRow.vue
Kaitou786: Thanks <3

flanksaul (approved):
Nice work!

Issue #123, opened 2021-02-17
Add restic check


moshloop:
Verify backup freshness, integrity and restorability

PR #158, opened 2021-04-11
feat: Add restic check support

Kaitou786:
Closes: #123


checks/restic.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+func (c *ResticChecker) Check(extConfig external.Check) *pkg.CheckResult {
+	logger.Debugf("Starting Restic checks")
+	start := time.Now()
+	resticCheck := extConfig.(v1.ResticCheck)
+	envVars := getEnvVars(resticCheck)
+	if err := checkIntegrity(resticCheck.Repository, resticCheck.CaCert, envVars); err != nil {
+		return Failf(resticCheck, "failed to check integrity of repository %v", err)
+	}
```
moshloop, line 46: This should be optional

checks/restic.go
```diff
@@ -0,0 +1,104 @@
...
+func (c *ResticChecker) Run(config v1.CanarySpec) []*pkg.CheckResult {
+	var results []*pkg.CheckResult
+	for _, conf := range config.Restic {
+		results = append(results, c.Check(conf))
+	}
+	return results
+}
+
+func (c *ResticChecker) Check(extConfig external.Check) *pkg.CheckResult {
+	logger.Debugf("Starting Restic checks")
```
moshloop, line 40:
```suggestion
```

checks/restic.go
```diff
@@ -0,0 +1,104 @@
...
+	resticCheck := extConfig.(v1.ResticCheck)
+	envVars := getEnvVars(resticCheck)
+	if err := checkIntegrity(resticCheck.Repository, resticCheck.CaCert, envVars); err != nil {
+		return Failf(resticCheck, "failed to check integrity of repository %v", err)
+	}
+	logger.Debugf("Restic integrity check succeeded for repository: %s", resticCheck.Repository)
+	if err := checkBackupFreshness(resticCheck.Repository, resticCheck.MaxAge, resticCheck.CaCert, envVars); err != nil {
+		return Failf(resticCheck, "failed while checking backup freshness: %v", err)
+	}
+	logger.Debugf("Restic backup freshness check succeeded for repository: %s", resticCheck.Repository)
```
moshloop, line 51: This type of logging should be handled in Success or in the outer function
```suggestion
```

checks/restic.go
```diff
@@ -0,0 +1,104 @@
...
+	logger.Debugf("Starting Restic checks")
+	start := time.Now()
+	resticCheck := extConfig.(v1.ResticCheck)
+	envVars := getEnvVars(resticCheck)
+	if err := checkIntegrity(resticCheck.Repository, resticCheck.CaCert, envVars); err != nil {
+		return Failf(resticCheck, "failed to check integrity of repository %v", err)
+	}
+	logger.Debugf("Restic integrity check succeeded for repository: %s", resticCheck.Repository)
+	if err := checkBackupFreshness(resticCheck.Repository, resticCheck.MaxAge, resticCheck.CaCert, envVars); err != nil {
+		return Failf(resticCheck, "failed while checking backup freshness: %v", err)
```
moshloop, line 49: This shouldn't return an error, but rather a message with how much older it is than than the maxage

checks/restic.go
```diff
@@ -0,0 +1,104 @@
...
+	return results
+}
+
+func (c *ResticChecker) Check(extConfig external.Check) *pkg.CheckResult {
+	logger.Debugf("Starting Restic checks")
+	start := time.Now()
+	resticCheck := extConfig.(v1.ResticCheck)
+	envVars := getEnvVars(resticCheck)
+	if err := checkIntegrity(resticCheck.Repository, resticCheck.CaCert, envVars); err != nil {
+		return Failf(resticCheck, "failed to check integrity of repository %v", err)
```
moshloop, line 45:
```suggestion
		return Failf(resticCheck, "integrity check failed: %v", err)
```

api/v1/checks.go
```diff
@@ -163,6 +163,35 @@ func (c S3BucketCheck) GetType() string {
...
+	// Password for the restic repository
+	Password string `yaml:"password" json:"password"`
+	// MaxAge for backup freshness
+	MaxAge string `yaml:"maxAge" json:"maxAge"`
+	// AccessKey access key id for connection with aws s3, minio, wasabi, alibaba oss
+	AccessKey string `yaml:"accessKey,omitempty" json:"accessKey,omitempty"`
+	// SecretKey secret access key for connection with aws s3, minio, wasabi, alibaba oss
+	SecretKey string `yaml:"secretKey,omitempty" json:"secretKey,omitempty"`
+	// CaCert path to the root cert. In case of self-signed certificates
+	CaCert string `yaml:"caCert,omitempty" json:"caCert,omitempty"`
```
moshloop, line 180: This should probably be InsecureTLSSkipVerify like in http checlk

checks/restic.go
Kaitou786: So like exposing a CRD attribute `checkIntegrity` if set to true will check this?

checks/restic.go
Kaitou786: Okay... I can include how much the backup is older than the maxAge but the canary should fail in that case, right?
So Only the  return message needs to be adjusted right?
The `Failf` remains here, if I got it correctly?

api/v1/checks.go
Kaitou786: Okay... though in restic skipping a [TLS/Verification is not possible](https://github.com/restic/restic/issues/2656) user needs to define the path of CaRoot cert. I took this field as same in helmCheck

checks/restic.go
moshloop: Correct

api/v1/checks.go
```diff
@@ -163,6 +163,37 @@ func (c S3BucketCheck) GetType() string {
...
+type ResticCheck struct {
+	// Description about the current restic canary check
+	Description string `yaml:"description,omitempty" json:"description,omitempty"`
+	// Repository The restic repository path eg: rest:https://user:pass@host:8000/ or rest:https://host:8000/ or s3:s3.amazonaws.com/bucket_name
+	Repository string `yaml:"repository" json:"repository"`
+	// Password for the restic repository
+	Password string `yaml:"password" json:"password"`
+	// MaxAge for backup freshness
+	MaxAge string `yaml:"maxAge" json:"maxAge"`
+	// CheckIntegrity when enabled will check the Integrity and consistency of the restic reposiotry
```
moshloop, line 175: Is this repository or the latest snapshot ?

Kaitou786: The MaxAge is for the latest snapshot.... the integrity check is for the whole repository

moshloop: We don't want to run an integrity check on the whole repo, as it can easily grow to TB's - Only on the most recent snapshot

PR #167, opened 2021-04-21
chore: fail CI incase of change in auto-generated files

Kaitou786:


Kaitou786:
Converting to draft should be merged after #170 is merged so they can be a single Job

PR #170, opened 2021-04-21
chore: add golangci-lint

Kaitou786:


.github/workflows/lint.yml
```diff
@@ -0,0 +1,17 @@
+name: Lint
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
```
flanksaul, line 7: Would propose that lint checks should be run for all pushes, if relevant file types change

.github/workflows/lint.yml
```diff
@@ -0,0 +1,17 @@
...
+  pull_request:
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+      - name: golangci-lint
```
flanksaul, line 16: Can this be scoped to changes only to *.go files?

.golangci.yml
```diff
@@ -0,0 +1,38 @@
+run:
+  # timeout for analysis, e.g. 30s, 5m, default is 1m
+  timeout: 20m
+
+linters:
+  # please, do not use `enable-all`: it's deprecated and will be removed soon.
+  # inverted configuration with `enable-all` and `disable` is not scalable during updates of golangci-lint
+  disable-all: true
+  enable:
```
flanksaul, line 9: Do we believe in all these options?

checks/namespace.go
```diff
@@ -156,7 +155,7 @@ func (c *NamespaceChecker) Check(extConfig external.Check) *pkg.CheckResult {
 	if _, err := pods.Create(context.TODO(), pod, metav1.CreateOptions{}); err != nil {
 		return unexpectedErrorf(check, err, "unable to create pod")
 	}
-	pod, err = c.WaitForPod(ns.Name, pod.Name, time.Millisecond*time.Duration(check.ScheduleTimeout), v1.PodRunning)
+	pod, _ = c.WaitForPod(ns.Name, pod.Name, time.Millisecond*time.Duration(check.ScheduleTimeout), v1.PodRunning)
```
flanksaul, line 158: Should this error not be handled?

checks/namespace.go
```diff
@@ -182,7 +181,6 @@ func (c *NamespaceChecker) Check(extConfig external.Check) *pkg.CheckResult {
 	deleteOk := true
 	deletion := NewTimer()
 	if err := pods.Delete(context.TODO(), pod.Name, metav1.DeleteOptions{}); err != nil {
-		deleteOk = false
```
flanksaul, line 185: Do we have tests to validate this change?

checks/namespace.go
```diff
@@ -393,13 +390,12 @@ func (c *NamespaceChecker) getHttp(url string, timeout int64, deadline time.Time
 		hardDeadline = softTimeoutDeadline
 	}
 
-	ctx, _ := context.WithDeadline(context.Background(), hardDeadline)
+	ctx, _ := context.WithDeadline(context.Background(), hardDeadline) // nolint: govet
```
flanksaul, line 393: Is this a permanent exclusion?

cmd/docs.go
```diff
@@ -298,18 +299,18 @@ func fieldId(typ ast.Expr) string {
 func fieldType(typ ast.Expr) string {
 	switch typ.(type) { // nolint: gosimple
 	case *ast.Ident:
-		return toLink(typ.(*ast.Ident).Name)
+		return toLink(typ.(*ast.Ident).Name) // nolint: gosimple
```
flanksaul, line 302: Should these additions be permanent?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,17 @@
+name: Lint
+on:
+  push:
+    tags:
+      - v*
```
BrendanGalloway, line 5: what is the motivation for running on tag?  tag does not change any of the repo content, so I'm not sure why we would need to test again

BrendanGalloway:
[nit] In future, I would split the commits that add the lint logic and that fix all the linting errors, just for the sake of clearer git history.

flanksaul:
Would it be possible to do a write up for IDE support to go alongside adding linters?

.github/workflows/lint.yml
Kaitou786: I can remove it. I followed the same pattern we had on `karina`

.github/workflows/lint.yml
Kaitou786: Okay I'll add to run only if changes in `**.go`

.golangci.yml
Kaitou786: Yeah these are the same options that we are using on all of our other repos

checks/namespace.go
```diff
@@ -143,7 +142,7 @@ func (c *NamespaceChecker) Check(extConfig external.Check) *pkg.CheckResult {
 		return unexpectedErrorf(check, err, "unable to create namespace")
 	}
 	defer func() {
-		c.Cleanup(ns)
+		c.Cleanup(ns) // nolint: errcheck
```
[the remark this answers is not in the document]
Kaitou786: Yes... so the specified type of lint will be ignored for this line

checks/namespace.go
Kaitou786: This is a flakey one... I tried to handle this error but the namespace test started to fail, even in case of error we want to proceed. and check further steps

checks/namespace.go
Kaitou786: In the e2e test we are running all the checks including this namespace one, also if you see the line below we are returning fields from the function without any reference to this var... The var is used when returning a success scenario

checks/namespace.go
Kaitou786: From this type of lint `govet` yeah

Kaitou786:
> Would it be possible to do a write up for IDE support to go alongside adding linters?

Not sure where would that go. Also the configuration is different for each IDE. We can plan to add a dev section on the repos describing about the environment. Like for example go version used in the projects, the linters and other things. I believe we can have a follow-up issue for that as this issue is for adding linter so I am not sure if it makes sense to write the whole dev-doc in this. But definitely I can write a dev-doc on a separate PR, wdyt?

.github/workflows/lint.yml
Kaitou786: I removed other branches than master because that was causing the same check to run twice once for your push on the branch and one for the PR that you created with that branch. So for me it made sense to run it once the PR is merged (push to master)
I think its good to run in case when multiple PRs are been worked on the same time and maybe single one doesn't bring the issue but altogether might

checks/namespace.go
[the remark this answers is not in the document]
flanksaul: I understand. What I'm wondering is if this change is so we can add the linter or if we actually don't want to errcheck here.

.github/workflows/lint.yml
flanksaul: @BrendanGalloway Any other file types linted with the checker?

.github/workflows/lint.yml
flanksaul: > > Would it be possible to do a write up for IDE support to go alongside adding linters?
> 
> Not sure where would that go. Also the configuration is different for each IDE. We can plan to add a dev section on the repos describing about the environment. Like for example go version used in the projects, the linters and other things. I believe we can have a follow-up issue for that as this issue is for adding linter so I am not sure if it makes sense to write the whole dev-doc in this. But definitely I can write a dev-doc on a separate PR, wdyt?

Would go wherever else we have setup guidelines. I think it could be a really valuable investment. Agree it would be different for each IDE but you could cover yours, for example. @moshloop @BrendanGalloway Do you think adding an IDE doc would be best done here or in follow up?

checks/namespace.go
flanksaul: What about just logging it?

flanksaul:
> Not sure where would that go. Also the configuration is different for each IDE. We can plan to add a dev section on the repos describing about the environment. Like for example go version used in the projects, the linters and other things. I believe we can have a follow-up issue for that as this issue is for adding linter so I am not sure if it makes sense to write the whole dev-doc in this. But definitely I can write a dev-doc on a separate PR, wdyt?

Top level comments thread so poorly. :/

Would go wherever else we have setup guidelines. I think it could be a really valuable investment. Agree it would be different for each IDE but you could cover yours, for example. @moshloop @BrendanGalloway Do you think adding an IDE doc would be best done here or in follow up?

.github/workflows/lint.yml
flanksaul: > I removed other branches than master because that was causing the same check to run twice once for your push on the branch and one for the PR that you created with that branch. So for me it made sense to run it once the PR is merged (push to master)
I think its good to run in case when multiple PRs are been worked on the same time and maybe single one doesn't bring the issue but altogether might

I've always just found the fast feedback useful for when I'm working on a branch. Given the low cost of linting, seems fine to me to run it again on PR. What does everyone else think? @BrendanGalloway @moshloop 

Would running on push for changes without branch scoping differentiate between normal push and PR?

BrendanGalloway: I would run it on PR but not on push if we only want to run once?  Everything should be entering master via PR, so it will get checked.

and golang-lint only lints go files as far as I know

.github/workflows/lint.yml
BrendanGalloway: from a build job?  We need to build on tag

BrendanGalloway:
goland enables lint by default AFAIK.  And I'm not sure IDE setup should be a per-project function, unless there's something specific that needs to be done to stage a working dev or test environment for that project that is different from general practice.  Otherwise you're going to end up with a lot of duplicated content if every repo needs an explanation on how to enable linters.  I think maybe a general dev-onboarding / Flanksource best practices document would be more appropriate

.github/workflows/lint.yml
flanksaul: From a developer feedback perspective, would it not be valuable to run before a developer gets as far as making a PR?

Kaitou786:
> goland enables lint by default AFAIK. And I'm not sure IDE setup should be a per-project function, unless there's something specific that needs to be done to stage a working dev or test environment for that project that is different from general practice. Otherwise you're going to end up with a lot of duplicated content if every repo needs an explanation on how to enable linters. I think maybe a general dev-onboarding / Flanksource best practices document would be more appropriate

+1000
I totally agree!!

flanksaul:
> goland enables lint by default AFAIK. And I'm not sure IDE setup should be a per-project function, unless there's something specific that needs to be done to stage a working dev or test environment for that project that is different from general practice. Otherwise you're going to end up with a lot of duplicated content if every repo needs an explanation on how to enable linters. I think maybe a general dev-onboarding / Flanksource best practices document would be more appropriate

@BrendanGalloway  Lacking that at the moment, could we not perhaps capture that work as a task?

checks/namespace.go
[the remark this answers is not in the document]
Kaitou786: Its added so we don't check for error in here.

.github/workflows/lint.yml
Kaitou786: From a developer perspective I would say the best would be run it on his local machine instead on a GH action for faster feedback

Kaitou786: @BrendanGalloway so we remove it from pushes to any branch and just let it be on PR?

.github/workflows/lint.yml
Kaitou786: Ah okay I see

.github/workflows/lint.yml
flanksaul: > From a developer perspective I would say the best would be run it on his local machine instead on a GH action for faster feedback

There is nothing stopping this but it requires manual action. CI is automated. The more automated things check my work the fewer I have to think about at any time.

I still think we should always lint on CI on changes at least.

checks/namespace.go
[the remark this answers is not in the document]
flanksaul: Because we don't want to know if there is an error?

.github/workflows/lint.yml
BrendanGalloway: For simple stuff like linting it's probably ok for it to run on push, if it's rerun on PR it's not a big resource consumption.  Was thinking more about the more resource intensive test setups we have

BrendanGalloway:
@moshloop where would an dev setup / onboarding / best practices guide live?  sre-resources or a new project?

checks/namespace.go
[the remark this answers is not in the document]
Kaitou786: Normally in defer functions/calls we won't be able to do much when error occurs (as we won't be able to return and use the value) So normally I have seen that we ignore lint checks in these cases, but if there is a better way do let me know :)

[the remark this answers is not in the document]
flanksaul: They can still be logged, right?

Otherwise one can use a callbacks to handle events at the end of async calls.  Check out this pattern in JS (because it's quick for me): https://codesandbox.io/s/mutable-tree-v5ron?file=/src/index.js:0-348

Usually on front end I'd have a function we can pipe stuff like this to and then decide what to do with it. Not sure what the go way is.

We probably would need to take a more global decisions. I don't like swallowing errors though.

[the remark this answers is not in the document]
flanksaul: @BrendanGalloway, views?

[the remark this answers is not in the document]
Kaitou786: I am not sure how to handle events at the end of async calls in golang, but I can log. Unless you guys know a better approach

moshloop:
Added https://github.com/flanksource/style-guide/issues/5 and https://github.com/flanksource/style-guide/issues/4

@BrendanGalloway dev setup for a specific project should be in a CONTRIBUTING.md  / general setup and onboarding probably in the knowledge base

PR #219, opened 2021-08-02
feat: improve junit check

Kaitou786:


checks/junit.go
```diff
@@ -70,22 +77,49 @@ func (c *JunitChecker) Check(extConfig external.Check) *pkg.CheckResult {
...
 	}
 	if junitCheck.GetName() != "" {
-		pod.Name = junitCheck.GetName()
+		pod.Name = junitCheck.GetName() + "-" + strconv.Itoa(int(start.Unix()))
 	} else {
 		name := rand.String(5)
 		pod.Name = strings.ToLower(name)
 	}
+	existingPods := getJunitPods(c.kommons, pod.Namespace)
+	if len(existingPods) != 0 {
```
moshloop, line 102: I don't think this if statement is needed

checks/junit.go
```diff
@@ -70,22 +77,49 @@ func (c *JunitChecker) Check(extConfig external.Check) *pkg.CheckResult {
...
+		junitCheckSelector: junitCheckLabelValue,
+	}
 	if junitCheck.GetNamespace() != "" {
 		pod.Namespace = junitCheck.GetNamespace()
 	} else {
 		pod.Namespace = corev1.NamespaceDefault
 	}
 	if junitCheck.GetName() != "" {
-		pod.Name = junitCheck.GetName()
+		pod.Name = junitCheck.GetName() + "-" + strconv.Itoa(int(start.Unix()))
```
moshloop, line 96: This will be an epoch - which isn't really useful, either use `yyyymmddhhmm` or random 4 - 6 chars

checks/junit.go
```diff
@@ -70,22 +77,49 @@ func (c *JunitChecker) Check(extConfig external.Check) *pkg.CheckResult {
...
 	}
+	existingPods := getJunitPods(c.kommons, pod.Namespace)
+	if len(existingPods) != 0 {
+		for _, junitPod := range existingPods {
+			obj, err := c.kommons.GetByKind(podKind, junitPod.Namespace, junitPod.Name)
+			if err != nil {
+				return junitFailF(junitCheck, textResults, junitStatus, template, "error fetching the pod: %v", err)
+			}
+			if obj != nil {
+				logger.Tracef("pod already exist")
```
moshloop, line 109:
```suggestion
```

checks/junit.go
```diff
@@ -200,3 +243,27 @@ func junitTemplateResult(template string, passed, failed, skipped, error int) (m
 	}
 	return message
 }
+
+func getJunitPods(kommonsClient *kommons.Client, namespace string) []corev1.Pod {
+	client, err := kommonsClient.GetClientset()
+	if err != nil {
+		return nil
+	}
+	//kommonsClient.GetFirstPodByLabelSelector()
```
moshloop, line 252:
```suggestion
```

checks/junit.go
```diff
@@ -97,17 +131,26 @@ func (c *JunitChecker) Check(extConfig external.Check) *pkg.CheckResult {
...
-	defer c.kommons.DeleteByKind(pod.Kind, pod.Namespace, pod.Name) // nolint: errcheck
+	defer c.kommons.DeleteByKind(podKind, pod.Namespace, pod.Name) // nolint: errcheck
 	logger.Tracef("waiting for pod to be ready")
-	err = c.kommons.WaitForPod(pod.Namespace, pod.Name, maxTime*time.Minute, corev1.PodRunning)
+	err = c.kommons.WaitForPod(pod.Namespace, pod.Name, time.Duration(timeout)*time.Minute, corev1.PodRunning)
 	if err != nil {
-		return junitFailF(junitCheck, textResults, junitStatus, template, "error waiting for pod: %v", err)
+		return junitFailF(junitCheck, textResults, junitStatus, template, "timeout waiting for pod: %v", err)
+	}
+	podObj, err := getJunitPod(c.kommons, pod.Namespace, pod.Name)
```
moshloop, line 147:
```suggestion
        var podObj v1.Pod
        err := c.kommons.Get(pod.Namespace, pod.Name, &podObj)
```

checks/junit.go
```diff
@@ -70,22 +77,49 @@ func (c *JunitChecker) Check(extConfig external.Check) *pkg.CheckResult {
...
 	pod := &corev1.Pod{}
 	pod.APIVersion = corev1.SchemeGroupVersion.Version
 	pod.Kind = podKind
+	pod.Labels = map[string]string{
+		junitCheckSelector: junitCheckLabelValue,
+	}
 	if junitCheck.GetNamespace() != "" {
 		pod.Namespace = junitCheck.GetNamespace()
 	} else {
 		pod.Namespace = corev1.NamespaceDefault
```
moshloop, line 93: Not sure how this would work? when would the namespace not be defined ?

checks/junit.go
```diff
@@ -70,22 +77,49 @@ func (c *JunitChecker) Check(extConfig external.Check) *pkg.CheckResult {
...
-		pod.Name = junitCheck.GetName()
+		pod.Name = junitCheck.GetName() + "-" + strconv.Itoa(int(start.Unix()))
 	} else {
 		name := rand.String(5)
 		pod.Name = strings.ToLower(name)
 	}
+	existingPods := getJunitPods(c.kommons, pod.Namespace)
+	if len(existingPods) != 0 {
+		for _, junitPod := range existingPods {
+			obj, err := c.kommons.GetByKind(podKind, junitPod.Namespace, junitPod.Name)
```
moshloop, line 104: Why do we need todo the lookup? we already have a ref `junitPod` ?

checks/junit.go
```diff
@@ -70,22 +77,49 @@ func (c *JunitChecker) Check(extConfig external.Check) *pkg.CheckResult {
...
+			if err != nil {
+				return junitFailF(junitCheck, textResults, junitStatus, template, "error fetching the pod: %v", err)
+			}
+			if obj != nil {
+				logger.Tracef("pod already exist")
+				createTime := obj.GetCreationTimestamp()
+				duration := time.Since(createTime.Time)
+				dur := uint64(duration.Seconds())
```
moshloop, line 112: Don't use abbreviations except for those that are extremely common

In this case I don't think you need the extra objects

```
time.Since(createTime.Time) > time.Seconds(interval * 2)

```

api/v1/checks.go
```diff
@@ -563,8 +563,13 @@ type JunitCheck struct {
 	// namespace in which canary is created
 	namespace string `yaml:"-" json:"-"`
 	// name of the canary. will be the same for the pod
-	name string     `yaml:"-" json:"-"`
-	Spec v1.PodSpec `yaml:"spec" json:"spec"`
+	name string `yaml:"-" json:"-"`
+	// interval of the canary
+	interval uint64 `yaml:"-" json:"-"`
+	//timeout in minutes to wait for specified container to finish its job.
```
moshloop, line 569: Maybe we just inject the canary back in ?

` parent Canary`

checks/junit.go
Kaitou786: When we are running once-off checks and don't pass the namespace flag so from the code, an empty string would be passed in that case.

api/v1/checks.go
Kaitou786: What do you mean?

PR #225, opened 2021-08-05
Authentication improvements

Kaitou786:


checks/docker.go
```diff
@@ -84,9 +92,22 @@ func (c *DockerPullChecker) Check(extConfig external.Check) *pkg.CheckResult {
...
+	if check.Auth != nil {
+		_, username, err = c.kommons.GetEnvValue(check.Auth.Username, namespace)
+		if err != nil {
+			return Failf(check, "failed to fetch username from envVar: %v", err)
+		}
+		_, password, err = c.kommons.GetEnvValue(check.Auth.Password, namespace)
+		if err != nil {
+			return Failf(check, "failed to fetch password from envVar: %v", err)
+		}
+	}
```
moshloop, line 107: I think we can extract this into a common method e.g. 

`auth, err := GetAuthValues(auth, check)`

moshloop:
[[CC] Refactor Auth](https://trello.com/c/DGrDMFnT/19-cc-refactor-auth)

checks/common.go
```diff
@@ -0,0 +1,21 @@
+package checks
+
+import (
+	v1 "github.com/flanksource/canary-checker/api/v1"
+	"github.com/flanksource/kommons"
+)
+
+func GetAuthValues(auth *v1.Authentication, client *kommons.Client, namespace string) (username, password string, err error) {
```
moshloop, line 8: I would return a new copy of Auth with the values filled in - This would allow future expansion to server/client-side TLS where supported etc...

checks/docker_push.go
```diff
@@ -35,9 +43,15 @@ func (c *DockerPushChecker) Check(extConfig external.Check) *pkg.CheckResult {
 	check := extConfig.(v1.DockerPushCheck)
 	start := time.Now()
 	ctx := context.Background()
+	namespace := check.GetNamespace()
+	var err error
+	check.Auth, err = GetAuthValues(check.Auth, c.kommons, namespace)
```
moshloop, line 48: We should create a new auth object, not overwrite the check as the credentials may then leak back into a patch etc..

Kaitou786: Changed

Issue #210, opened 2021-07-28
Add labels


moshloop:
Add labels from the canary spec into the API response + namespace and labels from the canary-checker pod prefixed with _

moshloop:
![](https://github.trello.services/images/mini-trello-icon.png) [[CC] Add Labels](https://trello.com/c/KDatLPj6/9-cc-add-labels)

PR #228, opened 2021-08-05
feature: add canary labels to API output

paddatrapper:
Fixes: #210


paddatrapper:
[[CC] Add Labels](https://trello.com/c/KDatLPj6/9-cc-add-labels)

pkg/cache/cache.go
```diff
@@ -59,16 +60,18 @@ func (c *cache) InitCheck(checks v1.Canary) {
...
-			Endpoint:    check.GetEndpoint(),
-			Interval:    checks.Spec.Interval,
-			Owner:       checks.Spec.Owner,
-			IconURL:     check.GetIconURL(),
-			Severity:    checks.Spec.Severity,
+			Type:         check.GetType(),
+			Name:         checks.ID(),
+			Namespace:    checks.Namespace,
+			Labels:       checks.ObjectMeta.Labels,
+			RunnerLabels: utils.GetLabels(),
```
moshloop, line 67: Utils generally are only for pure functions e.g. well defined input and output e.g. `utils.GetLabels("/etc/podinfo")` however due the fact that this is reading from disk we should do this at startup and cache the result e.g. 

`pkg/runner.go`
```golang
var Labels map[string]string = make(map[string]string

func LoadLabels(path string){}
```
and then in `cmd/operator.go`

```golang
runner.Labels = runner.LoadLabels("/etc/podinfo")
```

PR #233, opened 2021-08-09
feat: add iconURL on tha canary

Kaitou786:


moshloop:
@Kaitou786 can you check the tests ? The icon also needs to go into each check - maybe extract an inline `Description` object with description, name and iconUrl ?

Kaitou786:
[[CC] Add icon urls](https://trello.com/c/mof3AoLB/31-cc-add-icon-urls)

PR #235, opened 2021-08-12
feat: change run and check interface to enable canary object access i…

Kaitou786:
…nside every check


Kaitou786:
[[CC] Put Canary back into checks](https://trello.com/c/9RjSbSsA/53-cc-put-canary-back-into-checks)

PR #239, opened 2021-08-13
Ec2 checker

BrendanGalloway:
Move run logic from cmd to checks to avoid circular dependency
extend templating function to allow check defined input values
Add ec2 checker
limit number of ec2checkers to one per canary to avoid tag conflicts
allow nesting of canaries within an EC2 check


api/v1/canary_types.go
```diff
@@ -52,10 +52,12 @@ type CanarySpec struct {
 	Helm           []HelmCheck           `yaml:"helm,omitempty" json:"helm,omitempty"`
 	Namespace      []NamespaceCheck      `yaml:"namespace,omitempty" json:"namespace,omitempty"`
 	Redis          []RedisCheck          `yaml:"redis,omitempty" json:"redis,omitempty"`
-	IconURL        string                `yaml:"iconURL,omitempty" json:"iconURL,omitempty"`
-	Interval       uint64                `yaml:"interval,omitempty" json:"interval,omitempty"`
-	Severity       string                `yaml:"severity,omitempty" json:"severity,omitempty"`
-	Owner          string                `yaml:"owner,omitempty" json:"owner,omitempty"`
+	// Because the EC2 check attempts to clean up stale instances at the start of the check, only one EC2Check can be scoped per Canary instance
+	EC2      EC2Check `yaml:"ec2,omitempty" json:"ec2,omitempty"`
```
moshloop, line 56: It should only be cleaning up instances created by that check ?

api/v1/checks.go
```diff
@@ -958,6 +957,29 @@ type Smb struct {
 	SmbCheck `yaml:",inline" json:",inline"`
 }
 
+type EC2Check struct {
+	Description   `yaml:",inline" json:",inline"`
+	AccessKeyID   kommons.EnvVar            `yaml:"accessKeyID" json:"accessKeyID,omitempty"`
+	SecretKey     kommons.EnvVar            `yaml:"secretKey" json:"secretKey,omitempty"`
```
Kaitou786, line 963: Can there be any scenario when accessKey or SecretKey is not needed. If not I think we should make them as required fields. Also if they can be empty I think it would be better to use `*kommons.EnvVar` so before fetching the check we see if it is empty or not

checks/ec2.go
```diff
@@ -0,0 +1,415 @@
...
+		return HandleFail(check, fmt.Sprintf("Could not parse EC2 secret key: %v", err))
+	}
+	tr := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: false},
+	}
+	if check.SkipTLSVerify {
+		tr = &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		}
+	}
```
Kaitou786, line 132:
```suggestion
	tr := &http.Transport{
		TLSClientConfig: &tls.Config{InsecureSkipVerify: check.SkipTLSVerify},
	}
```

checks/ec2.go
```diff
@@ -0,0 +1,415 @@
...
+	kommons := c.GetClient()
+	if kommons == nil {
+		return HandleFail(check, "Kommons client not configured for ec2 checker")
+	}
+	_, ak, err := kommons.GetEnvValue(check.AccessKeyID, namespace)
+	if err != nil {
+		return HandleFail(check, fmt.Sprintf("Could not parse EC2 access key: %v", err))
+	}
+	_, sk, err := kommons.GetEnvValue(check.SecretKey, namespace)
+	if err != nil {
```
Kaitou786, line 122: nitpick: I don't think we should be using abbreviation `ak` and `sk` here. What's wrong with using accessKey and secretKey ?

checks/ec2.go
```diff
@@ -0,0 +1,415 @@
...
+	cfg, err := config.LoadDefaultConfig(context.TODO(),
+		config.WithCredentialsProvider(credentials.NewStaticCredentialsProvider(ak, sk, "")),
+		config.WithRegion(check.Region),
+		config.WithHTTPClient(&http.Client{Transport: tr}),
+	)
+
+	var ami *string
+	if check.AMI == "" {
+		ssmClient := ssm.NewFromConfig(cfg)
+		arnLookupInput := &ssm.GetParameterInput{Name: aws.String("/aws/service/ami-amazon-linux-latest/amzn2-ami-hvm-x86_64-gp2")}
```
Kaitou786, line 142: The ami is hardcoded?

checks/ec2.go
```diff
@@ -0,0 +1,415 @@
...
+				Name: aws.String("tag:owner"),
+				Values: []string{
+					idString,
+				},
+			},
+			{
+				Name: aws.String("instance-state-name"),
+				Values: []string{
+					"running",
+				},
```
Kaitou786, line 173: I might be missing something but In case an instance is pending or any other state than running for a long time shouldn't we fail the test instead of terminating it. Maybe we wait for it to be running for a specific time period if it is not running until then we fail and start to create a new instance?

checks/ec2.go
```diff
@@ -0,0 +1,415 @@
...
+	ip := ""
+	internalDNS := ""
+
+	logger.Infof("Created EC2 instance with id %v", *instanceId)
+	var startTime float64
+	for {
+		describeInput := &ec2.DescribeInstancesInput{InstanceIds: []string{*instanceId}}
+		describeOutput, err := client.DescribeInstances(context.TODO(), describeInput)
+		if err != nil {
+			return &pkg.CheckResult{ // nolint: staticcheck
```
Kaitou786, line 251: Why not use handleFail here as well?

checks/ec2.go
```diff
@@ -0,0 +1,415 @@
...
+			break
+		}
+		if timer.Millis() > 300000 {
+			return HandleFail(check, fmt.Sprintf("Instance did not start within 5 minutes: %v", *reason.Message))
+		}
+		time.Sleep(1 * time.Second)
+	}
+	prometheusStartupTime.WithLabelValues(check.Region).Set(startTime)
+	time.Sleep(time.Duration(check.WaitTime) * time.Second)
+	fmt.Println(ip)
```
Kaitou786, line 281:
```suggestion
```

Kaitou786 (commented):
The check method for ec2 seems way to long, do you think we can factor that more into smaller logical functions?

api/v1/canary_types.go
BrendanGalloway: the instances are tagged with the name of the Canary, so if there are multiple checks per canary, there isn't a unique identifier to tell which check within the canary is which.

checks/ec2.go
BrendanGalloway: this arn is an amazon provided ssm parameter that always points to the latest amazon linux ami.  I'm using it as a fallthrough if no ami is specified in the check config.

checks/ec2.go
BrendanGalloway: this is a pre-run cleanup that terminates instances in the case that a previous run created an instance and then failed to terminate them.  Previously created pending instances should probably also be cleaned up.

api/v1/checks.go
BrendanGalloway: You have to log into aws to perform this check, so the credentials will always need to be provided.

Kaitou786: Yeah so in the when both are required fields I suggest we remove the `omitempty` that way user would have to provide them before the CR can be created

checks/ec2.go
Kaitou786: In that case can we extract it to a constant?

checks/ec2.go
Kaitou786: yeah they need to be cleaned for sure... but I think we should be failing in case the instance is running for a long time or if it is stuck in any other state I would terminate it and fail the test
@moshloop wdyt?

BrendanGalloway: That is what I am doing?  If the instances do not start or stop within five minutes then the test fails and exits.  The initial snippet is to handle the case where that has happened in the past and now there instances from a previous test that are still running.  The test will fail again should terminating these instances fail before a new instance is started.

moshloop:
Merging to get the API in

checks/ec2.go
Kaitou786: Yeah... but that is hard-coded 5 minutes... maybe that can be more flexible depending on the particular ec2Check?
Like in Junit we wait until 2*interval after that we terminate and delete. Also, can we extract them as constants as well?

BrendanGalloway: I have added a timeout input

PR #246, opened 2021-08-17
ci: move aws tests to cron pipeline

BrendanGalloway:


Kaitou786:
@BrendanGalloway seems there is some error in the workflow file, the action didn't ran and we check the error from: https://github.com/flanksource/canary-checker/actions/runs/1138923854

PR #276, opened 2021-08-25
feat: add gcsbucket support

Kaitou786:


api/v1/checks.go
```diff
@@ -139,6 +139,60 @@ func (c S3BucketCheck) GetType() string {
 	return "s3Bucket"
 }
 
+type GCPConnection struct {
+	Endpoint        string          `yaml:"endpoint" json:"endpoint,omitempty"`
+	CredentialsJSON *kommons.EnvVar `yaml:"credentialsJSON" json:"credentialsJSON,omitempty"`
```
moshloop, line 144:
```suggestion
	Credentials *kommons.EnvVar `yaml:"credentials" json:"credentials,omitempty"`
```

checks/gcs_bucket.go
```diff
@@ -0,0 +1,130 @@
...
+		return result.ErrorMessage(err)
+	}
+	result.AddDetails(folders)
+	if test := folders.Test(check.FolderTest); test != "" {
+		result.Failf(test)
+	}
+	return result
+}
+
+func (conn *GCS) CheckFolder(ctx *context.Context, query *v1.GCSQuery) (*FolderCheck, error) {
```
moshloop, line 89: Can't we reuse the function from smb, seeing as we are getting a GCSFileinfo ?

api/v1/checks.go
```diff
@@ -139,6 +139,60 @@ func (c S3BucketCheck) GetType() string {
...
+	// listed will have names between startOffset (inclusive) and endOffset (exclusive).
+	EndOffset string `yaml:"endOffset,omitempty" json:"endOffset,omitempty"`
+}
+
+type GCSBucketCheck struct {
+	Description   `yaml:",inline" json:",inline"`
+	Templatable   `yaml:",inline" json:",inline"`
+	FolderTest    `yaml:",inline" json:",inline"`
+	GCPConnection `yaml:",inline" json:",inline"`
+	*GCSQuery     `yaml:",inline" json:",inline"`
```
moshloop, line 184: I think we only want a path (a regex for matching the filename)

The prefix should be derived from the bucket (id specified)

checks/gcs_bucket.go
```diff
@@ -0,0 +1,130 @@
...
+	result := FolderCheck{}
+	var objs *gcs.ObjectIterator
+	bucket := conn.Bucket(conn.BucketName)
+	if query != nil {
+		objs = bucket.Objects(ctx, &gcs.Query{
+			Prefix:      query.Prefix,
+			Versions:    query.Versions,
+			Delimiter:   query.Delimiter,
+			StartOffset: query.StartOffset,
+			EndOffset:   query.EndOffset,
```
moshloop, line 99: These shouldn't be specified by a user - we need to page through all the results.

checks/gcs_bucket.go
```diff
@@ -0,0 +1,130 @@
...
+func (c *GCSBucketChecker) Type() string {
+	return "gcsBucket"
+}
+
+type GCS struct {
+	BucketName string
+	*gcs.Client
+}
+
+type GCSFileInfo struct {
```
moshloop, line 39: I would move this struct to pkg/gcp

PR #277, opened 2021-08-25
fix: dangling checks in case check key is changed

Kaitou786:


Kaitou786:
[[CC] Dangling Checks on the uI](https://trello.com/c/X0z4Oosv/74-cc-dangling-checks-on-the-ui)

api/v1/canary_types.go
```diff
@@ -203,11 +213,10 @@ type CanaryStatus struct {
 // Canary is the Schema for the canaries API
 // +kubebuilder:printcolumn:name="Interval",type=string,JSONPath=`.spec.interval`
 // +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.status`
-// +kubebuilder:printcolumn:name="Message",type=string,JSONPath=`.status.message`
+// +kubebuilder:printcolumn:name="Last Check",type=date,JSONPath=`.status.lastCheck`
```
moshloop, line 216: I think we still want the (first/error) message ?

moshloop: I think we take the first non-empty message/error and then add `...` if there are more

PR #319, opened 2021-09-06
CI fixes

Kaitou786:


moshloop:
@Kaitou786  can we move helm to datasources ?

.github/workflows/test.yml
```diff
@@ -48,6 +48,10 @@ jobs:
           restore-keys: |
             cache-
       - run: make bin
+      - name: Install restic
+        run: |
+          sudo apt-get install restic
```
moshloop, line 53: were we not going to move this to commons.Dep ?

Kaitou786: While checking for backup freshness in restic we run restic command inside restic command [example](https://github.com/flanksource/canary-checker/blob/master/checks/restic.go#L78)
I am not sure how to do that with the common.Deps.... also since inside the container this problem doesn't arise as we add the binary in the container image and running the e2e test on the local system the path is preserved. So installing restic here seemed like a valid solution

PR #345, opened 2021-10-04
feat: add endpoint to retrieve check. Closes #242

### LINKED ISSUE #242: /api/spec/key
Retrieve the check YAML for a given check

Kaitou786:


Kaitou786:
[[CC] Retrieve Yaml endpoint](https://trello.com/c/8kEcZzjW/95-cc-retrieve-yaml-endpoint)

pkg/api.go
```diff
@@ -92,7 +92,10 @@ type Check struct {
 	Severity     string            `json:"severity"`
 	Icon         string            `json:"icon"`
 	DisplayType  string            `json:"displayType"`
-	CheckCanary  *v1.Canary        `json:"-"`
+	RunnerName   string            `json:"runnerName"`
+	// Specify the canary location, <runner>/<namespace>/<name>
+	Location string     `json:"location"`
```
moshloop, line 97: What is the different between runner name and location ?

Kaitou786: So the runner name is just the name of the runner. The Location would be <runner-name>/<namespace>/<name>. I added this so we don't end up with nested loops to fetch the canary object, with these parameters

PR #347, opened 2021-10-05
Service installer

Kaitou786:
Tested to be working on the MacOS with the push servers and other flags... need to verify the same for a windows instance.


cmd/install_service.go
```diff
@@ -0,0 +1,113 @@
...
+	"path/filepath"
+
+	"github.com/flanksource/canary-checker/pkg/cache"
+	"github.com/flanksource/canary-checker/pkg/runner"
+	"github.com/kardianos/service"
+	"github.com/spf13/cobra"
+)
+
+var InstallService = &cobra.Command{
+	Use:   "install-service",
```
moshloop, line 15:
```suggestion
	Use:   "install",
```

PR #355, opened 2021-10-13
feat: add exec check

Kaitou786:


moshloop:
@Kaitou786  can you add fixtures and update the docs ?

PR #356, opened 2021-10-14
feat: add cache chain with postgresql support

Kaitou786:


cmd/root.go
```diff
@@ -36,6 +36,11 @@ func ServerFlags(flags *pflag.FlagSet) {
 	flags.StringSliceVar(&pullServers, "pull-servers", []string{}, "push check results to multiple canary servers")
 	flags.StringVar(&runner.RunnerName, "name", "local", "Server name shown in aggregate dashboard")
 	flags.StringVar(&prometheusURL, "prometheus", "", "URL of the prometheus server that is scraping this instance")
+	flags.StringVar(&cache.PostgresHost, "postgres-host", "", "Address for the postgres database")
```
moshloop, line 39: I think we just want to use a connection string with a default of CACHE_CONNECTION_STRING env var

pkg/api/api.go
```diff
@@ -18,9 +20,38 @@ type Response struct {
...
+			return
+		}
+		timeDuration = &duration
+	}
+	checkKey := queryParams.Get("key")
+
 	apiResponse := &Response{
 		RunnerName: runner.RunnerName,
-		Checks:     cache.GetChecks(""),
+		Checks:     cache.QueryChecks(cache.CacheChain, c, timeDuration, checkKey),
```
moshloop, line 54: I don't think we want to expose CacheChain to the outside world - it should be private and the we have package level methods that delegate to it

pkg/cache/postgres.go
```diff
@@ -0,0 +1,348 @@
...
+		key TEXT NOT NULL,
+		checktype TEXT NOT NULL,
+		name TEXT NOT NULL,
+		namespace TEXT NOT NULL,
+		labels json,
+		runnerLabels json,
+		canaryName TEXT,
+		description TEXT,
+		endpoint TEXT,
+		Interval int,
```
moshloop, line 45: sort the field name alphabetically

moshloop, line 45:
```suggestion
		interval int,
```

pkg/cache/postgres.go
```diff
@@ -0,0 +1,348 @@
...
+
+func InitPostgres(username, password, database, host string, port int) (*pgxpool.Pool, error) {
+	connString := fmt.Sprintf("postgres://%v:%v@%v:%d/%v", username, password, host, port, database)
+	conn, err := pgxpool.Connect(context.Background(), connString)
+	if err != nil {
+		return nil, err
+	}
+	_, err = conn.Exec(context.TODO(), `CREATE TABLE IF NOT EXISTS checks(
+		key TEXT NOT NULL,
+		checktype TEXT NOT NULL,
```
moshloop, line 37:
```suggestion
		checkType TEXT NOT NULL,
```

or 

```suggestion
		check_type TEXT NOT NULL,
```

pkg/cache/postgres.go
```diff
@@ -0,0 +1,348 @@
...
+		CheckKey TEXT NOT NULL,
+		inserted_at TIMESTAMP NOT NULL,
+		PRIMARY KEY (time, CheckKey)
+	)`)
+	if err != nil {
+		return nil, err
+	}
+	cron := cron.New()
+	cron.AddFunc("@every 1d", func() { // nolint: errcheck
+		if _, err := conn.Exec(context.TODO(), "DELETE FROM checks WHERE updated_at < NOW() - INTERVAL '28 days';"); err != nil {
```
moshloop, line 78: The interval should be made a command line argument with a default of 90 days

pkg/cache/postgres.go
```diff
@@ -0,0 +1,348 @@
...
+	if _, err := c.Conn.Exec(context.TODO(), `DELETE FROM check_statuses WHERE checkKey=$1`, key); err != nil {
+		logger.Errorf("error deleting check_statuses from postgres: %v", err)
+	}
+}
+
+func (c *postgresCache) ListCheckStatus(checkKey string, count int64, duration *time.Duration) []pkg.CheckStatus {
+	if duration != nil {
+		seconds := fmt.Sprintf("%v SECOND", duration.Seconds())
+		// statusRows, err := c.Conn.Query(context.TODO(), `SELECT * FROM check_statuses WHERE checkKey=$1 and (inserted_at > NOW() - Interval $2) ORDER BY inserted_at DESC LIMIT $3`, checkKey, seconds, count)
+		statusRows, err := c.Conn.Query(context.TODO(), fmt.Sprintf(`SELECT * FROM check_statuses WHERE checkKey=$1 and (inserted_at > NOW() - Interval '%v') ORDER BY inserted_at DESC LIMIT $2`, seconds), checkKey, count)
```
moshloop, line 301: Can we add a FIXME and a description of why the first option is not used ?

pkg/changes/changes.go
```diff
@@ -38,7 +38,7 @@ func Handler(w http.ResponseWriter, req *http.Request) {
 
 	results := []changes.Changes{}
 
-	for _, check := range cache.Cache.Checks {
+	for _, check := range cache.InMemoryCache.Checks {
```
moshloop, line 41: This should be via cache.GetChecks() which delegates to the chain

PR #394, opened 2021-12-16
feat: add canary-checker push sub-command

Kaitou786:


cmd/push.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/flanksource/canary-checker/pkg/cache"
+	"github.com/flanksource/canary-checker/pkg/push"
+	"github.com/flanksource/commons/logger"
+	"github.com/spf13/cobra"
+)
+
+var name, description, checkType, status, duration, message, checkDetails, errorMessage string
+var Push = &cobra.Command{
+	Use:   "push",
+	Short: "Push a check to the canary-checker-server",
```
moshloop, line 21:
```suggestion
	Use:   "push servers",
	Short: "Push a check to the canary-checker-server",
```

cmd/push.go
```diff
@@ -0,0 +1,97 @@
...
+				} else {
+					check.Uptime.Failed += 1
+				}
+			} else {
+				if passed {
+					check.Uptime.Passed = 1
+				} else {
+					check.Uptime.Failed = 1
+				}
+			}
```
moshloop, line 71: I don't see how this will ever be populated ?

moshloop (changes requested):

cmd/push.go
Kaitou786: Yeah it would never be executed

PR #410, opened 2022-02-02
Awsconfigrules

Kaitou786:


api/v1/checks.go
```diff
@@ -704,6 +705,24 @@ func (c AwsConfigCheck) GetEndpoint() string {
 	return c.Query
 }
 
+type AwsConfigRuleCheck struct {
+	Description `yaml:",inline" json:",inline"`
+	Templatable `yaml:",inline" json:",inline"`
+	// Specify one or more Config rule names to filter the results by rule.
+	Rules []string `yaml:"rules,omitempty" json:"rules,omitempty"`
```
moshloop, line 712: I think we want to invert these in go

Kaitou786: added `ignoreRules` for that

PR #417, opened 2022-02-28
Topology and switch to gorm

moshloop:


Kaitou786:
It seems the operator doesn't start with these changes.

PR #421, opened 2022-03-09
Refactor helm chart and get it running

paddatrapper:


paddatrapper:
This now creates the DB, installs the CRDs and canary-checker runs correctly

PR #428, opened 2022-03-17
fix: initialise DB in operator command

paddatrapper:


paddatrapper:
This seems not to be necessary, as `serve` is called from the operator

PR #441, opened 2022-03-28
feat: add database backup check

BrendanGalloway:
First pass at creating a canary to verify that database backups have occured.
Just caters for GCP now, but can be expanded to cover other APIs too.
Probably missing bunch of the newer canary functionality, so input would be appreciated.


checks/database_backup.go
```diff
@@ -0,0 +1,65 @@
...
+	for _, conf := range ctx.Canary.Spec.DatabaseBackup {
+		results = append(results, c.Check(ctx, conf)...)
+	}
+	return results
+}
+
+func (c *DatabaseBackupChecker) Check(ctx *context.Context, extConfig external.Check) pkg.Results {
+	check := extConfig.(v1.DatabaseBackupCheck)
+	switch {
+	case check.GCP.Instance != "" && check.GCP.Project != "":
```
Kaitou786, line 52: If both of these are mandatory fields I would remove the `omitempty` from the yaml annotation

Kaitou786: And then switch on the case `check.GCP != nil` by making GCP as a pointer reference

checks/database_backup_gcp.go
```diff
@@ -0,0 +1,56 @@
...
+		databaseScanFailCount.WithLabelValues(check.GCP.Project, check.GCP.Instance).Inc()
+		return results.ErrorMessage(err)
+	}
+	var errorMessages []string
+	for _, backup := range backupList.Items {
+		if backup.Status != "SUCCESSFUL" {
+			errorMessages = append(errorMessages, backup.Error.Message)
+		}
+	}
+	if check.MaxAge > 0 {
```
Kaitou786, line 39: Can MaxAge of `0` be a valid case?
In that case I would also make it a pointer and have it as `check.MaxAge != nil`

checks/database_backup_gcp.go
```diff
@@ -0,0 +1,56 @@
...
+	// Only checking one backup for now, but setting up the logic that this could maybe be configurable.
+	// Would need some extra parsing on the age to select latest
+	backupList, err := svc.BackupRuns.List(check.GCP.Project, check.GCP.Instance).MaxResults(1).Do()
+	if err != nil {
+		databaseScanFailCount.WithLabelValues(check.GCP.Project, check.GCP.Instance).Inc()
+		return results.ErrorMessage(err)
+	}
+	var errorMessages []string
+	for _, backup := range backupList.Items {
+		if backup.Status != "SUCCESSFUL" {
```
Kaitou786, line 35: Maybe make it a constant... and not sure but could it also be `PENDING` and if so I would add the `backup.Status` in the errorMessages. Something like `Expecting backupStatus to be `SUCCESSFUL` but is backup.Status`

checks/database_backup_gcp.go
```diff
@@ -0,0 +1,56 @@
...
+			errorMessages = append(errorMessages, backup.Error.Message)
```
Kaitou786, line 36: Maybe add a `\n` char at the en of each append?

Kaitou786:
@BrendanGalloway can you also check the lint issues?

checks/database_backup_gcp.go
BrendanGalloway: Good point about in progress backups, I will need to find out what the status is for backupRuns that are in progress

checks/database_backup_gcp.go
BrendanGalloway: I think the \n characters get logged as \n and not newlines? If they don't, multiple newlines in a single log entry messes with most of the k8s logging tools unless you add special handling

Kaitou786: We do need a way to separate multiple messages in case backupList.Items have multiple backups... since the current loop will have the output in messed up state?

checks/database_backup_gcp.go
Kaitou786: There could be constants in the GCP package itself for the backup status

checks/database_backup_gcp.go
BrendanGalloway: I don't think it would be a reasonable case to always expect the most recent backup to be under a minute old...The most frequent backup schedule we would currently be monitoring is daily.

checks/database_backup_gcp.go
BrendanGalloway: I have used the upstream const

moshloop:
@BrendanGalloway can you rebase ?

checks/database_backup_gcp.go
```diff
@@ -32,18 +34,42 @@ func GCPDatabaseBackupCheck(ctx *context.Context, check v1.DatabaseBackupCheck)
 	}
 	var errorMessages []string
 	for _, backup := range backupList.Items {
-		if backup.Status != "SUCCESSFUL" {
-			errorMessages = append(errorMessages, backup.Error.Message)
+		if !(backup.Status == sql.SqlBackupRunStatus_SUCCESSFUL.String() || backup.Status == sql.SqlBackupRunStatus_RUNNING.String() || backup.Status == sql.SqlBackupRunStatus_ENQUEUED.String()) {
```
Kaitou786, line 37: I would rather make it a const list of allowedStatuses that makes it more readable and extendable in the future.

checks/database_backup_gcp.go
```diff
@@ -32,18 +34,42 @@ func GCPDatabaseBackupCheck(ctx *context.Context, check v1.DatabaseBackupCheck)
...
+				checkString = "enqueued"
 			} else {
-				if endTime.Add(time.Duration(check.MaxAge) * time.Minute).After(time.Now()) {
-					errorMessages = append(errorMessages, fmt.Sprintf("Most recent backup run too old - ended at %s", backup.EndTime))
+				checkTime, err = time.Parse(time.RFC3339, backup.EndTime)
+				if err != nil {
+					errorMessages = append(errorMessages, "Could not parse backup end time")
+					parseFail = true
+				}
+				checkString = "ended"
```
Kaitou786, line 68: Maybe `succeeded` ?

api/v1/checks.go
```diff
@@ -966,6 +987,13 @@ type AwsConfigRule struct {
 	AwsConfigRuleCheck `yaml:",inline" json:",inline"`
 }
 
+/*
+[include:datasources/database_backup.yaml]
```
Kaitou786, line 991: Maybe forgot to add this sample canary?

checks/database_backup_gcp.go
BrendanGalloway: I'm not specifically checking that the status is succeeded when doing this check, just assuming that the other states have an endtime.  Maybe I should only look at success

checks/database_backup_gcp.go
BrendanGalloway: const cannot be slices

Kaitou786: Then I would make it a var slice

checks/database_backup_gcp.go
BrendanGalloway: reworked this to check which time fields exist directly, rather than extrapolating from status

checks/database_backup_gcp.go
BrendanGalloway: They are separated by ', ' rather than '\n', see the return statement

fixtures/datasources/database_backup.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: canaries.flanksource.com/v1
+kind: Canary
+metadata:
+  name: database-backup-example
+spec:
+  interval: 60
+  databaseBackup:
+    - maxAge: 3600
```
Kaitou786, line 8: Apologies for noticing it this late in the review... but I think we need to take `string` kind here of type duration and parse it to convert it to seconds by code.... e.g: `5m30s`

PR #448, opened 2022-04-06
feat: add config db query check

Kaitou786:


checks/config_db.go
```diff
@@ -0,0 +1,57 @@
...
+	result := pkg.Success(check, ctx.Canary)
+	var results pkg.Results
+	results = append(results, result)
+	logger.Infof("Sending GET request to: %v with query param: %v", check.Host, check.Query)
+	var url string
+	if strings.HasSuffix(check.Host, "/") {
+		url = fmt.Sprintf("%v?query=%v", check.Host, check.Query)
+	} else {
+		url = fmt.Sprintf("%v/?query=%v", check.Host, check.Query)
+	}
```
Kaitou786, line 41: @moshloop  let me know if the endpoint needs to be updated

checks/config_db.go
```diff
@@ -0,0 +1,57 @@
...
+	return results
+}
+
+func (c *ConfigdbChecker) Check(ctx *canaryContext.Context, extConfig external.Check) pkg.Results {
+	check := extConfig.(v1.ConfigDBCheck)
+	result := pkg.Success(check, ctx.Canary)
+	var results pkg.Results
+	results = append(results, result)
+	logger.Infof("Sending GET request to: %v with query param: %v", check.Host, check.Query)
```
moshloop, line 36:
```suggestion
	logger.Tracef("[%v] query: %v", check.Host, check.Query)
```

checks/config_db.go
```diff
@@ -0,0 +1,57 @@
...
+	var url string
+	if strings.HasSuffix(check.Host, "/") {
+		url = fmt.Sprintf("%v?query=%v", check.Host, check.Query)
```
moshloop: We might need to url encode the query

Kaitou786: Added the QueryEscape parameter for query

checks/config_db.go
```diff
@@ -0,0 +1,57 @@
...
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusAccepted && resp.StatusCode != http.StatusCreated {
+		return results.Failf("Failed to get response, %v", resp.Status)
+	}
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return results.Failf("Failed to read the response body: %v", err)
+	}
+
+	result.AddDetails(body)
```
moshloop, line 57: Can we unmarshal the json into https://github.com/flanksource/config-db/blob/main/api/v1/interface.go#L41 ?

We can copy and pate the struct definition.

PR #450, opened 2022-04-07
fix: run as unpriveleged user

paddatrapper:


chart/templates/deployment.yaml
```diff
@@ -15,6 +15,7 @@ spec:
       {{- include "canary-checker.selectorLabels" . | nindent 6 }}
   {{- if eq .Values.db.external.enabled false }}
   {{- if eq .Values.db.embedded.persist true }}
+  serviceName: canary-checker
```
moshloop, line 18: Shouldn't this be: ` {{ include "canary-checker.fullname" . }}-sa` ?

paddatrapper: It should be `{{ include "canary-checker.fullname" . }}`, it is the DNS serviceName not the serviceAccountName. I've updated it

PR #469, opened 2022-04-27
System/Topology controller

Kaitou786:


Kaitou786:
Seems the failed test case is because of the new kommons update with the `isReady` logic: https://github.com/flanksource/kommons/pull/80

@BrendanGalloway can you please check what was the original problem for changing that logic and possibly fix it... WIth the current flow, I don't see any scenario where it would return `true` for the isRead since conditions is a list and the first one would in most cases be of 
```
 status: "True"
 type: Initialized
 ```
 which will in turn return false due to 
 ```
 if condition["type"] != "Ready" && condition["status"] != "False" {
			return false, fmt.Sprintf("⏳ waiting for %s/%s: %s", condition["type"], condition["status"], condition["message"])
		}
```

moshloop:
> Initialized

I think we might need to add this as a special case - Most conditions are falsey  when true it is in a non-ready state

Kaitou786:
> > Initialized
> 
> I think we might need to add this as a special case - Most conditions are falsey when true it is in a non-ready state

Shouldn't just making the condition as != True work?

pkg/system_api.go
```diff
@@ -91,36 +147,37 @@ func (components *Components) UnmarshalJSON(b []byte) error {
...
-	Lifecycle     string             `json:"lifecycle,omitempty"`
-	Relationships []RelationshipSpec `json:"relationships,omitempty"`
-	Properties    Properties         `json:"properties,omitempty"`
-	Components    Components         `json:"components,omitempty"`
-	ParentId      string             `json:"parent_id,omitempty"` //nolint
-	CreatedAt     string             `json:"created_at,omitempty"`
-	UpdatedAt     string             `json:"updated_at,omitempty"`
-	ExternalId    string             `json:"external_id,omitempty"` //nolint
+	Lifecycle     string        `json:"lifecycle,omitempty"`
+	Relationships Relationships `json:"relationships,omitempty" gorm:"type:relationships"`
```
moshloop, line 166: We can't embed this as a json into component - It will cause very high locking / deadlocking on the row when multiple components needed to be added, it will also make querying difficult as the component Id's will need to be extracted from the JSON . 

We would need s seperate table e.g. 

parentId 
compononentId
type
createdAt
systemTemplateId
updatedAt (Every topology run should refresh this, so that stale relationships can be removed)

- We should however move this to a subsequent PR.

pkg/db/migrations/3_topology.sql
```diff
@@ -2,55 +2,77 @@
 -- +goose StatementBegin
 
 
-CREATE TABLE system (
+
+CREATE TABLE IF NOT EXISTS SYSTEM_TEMPLATES (
```
moshloop, line 6: s/SYSTEM_TEMPLATES/system_templates/

This script should never run when the table already exists so the IF NOT EXISTS can come out. 
we also should not be making changes to this file, but rather adding additional migration scripts that alert the tables originally created here

pkg/controllers/system_controller.go
```diff
@@ -0,0 +1,79 @@
...
+		logger.V(1).Info("System not found")
+		return ctrl.Result{}, nil
+	}
+	// db.AddSystem(system)
+	id, changed, err := db.AddSystemTemplate(systemTemplate)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	if changed || systemTemplate.Status.ObservedGeneration == 1 {
+		SyncSystemsJobs()
```
moshloop, line 56: I think this needs to change to SyncSystemJob(systemTemplate)

pkg/controllers/canary_controller.go
```diff
@@ -100,7 +100,9 @@ func (r *CanaryReconciler) Reconcile(ctx gocontext.Context, req ctrl.Request) (c
 	canary.Status.PersistedID = &id
 	// Sync jobs if canary is created or updated
 	if canary.Generation == 1 || changed {
-		SyncCanaryJob(*canary)
+		if err := SyncCanaryJob(*canary); err != nil {
+			logger.Error(err, "failed to sync canary job")
```
moshloop, line 104: shouldn't this return a reconcile error ?

pkg/controllers/system_controller.go
```diff
@@ -0,0 +1,79 @@
...
+// +kubebuilder:rbac:groups=canaries.flanksource.com,resources=systemtemplates/status,verbs=get;update;patch
+func (r *SystemReconciler) Reconcile(ctx gocontext.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := r.Log.WithValues("system", req.NamespacedName)
+	systemTemplate := &v1.SystemTemplate{}
+	err := r.Get(ctx, req.NamespacedName, systemTemplate)
+	if errors.IsNotFound(err) {
+		logger.V(1).Info("System not found")
+		return ctrl.Result{}, nil
+	}
+	// db.AddSystem(system)
```
moshloop, line 50:
```suggestion
```

pkg/db/migrations/3_topology.sql
```diff
@@ -2,55 +2,77 @@
...
   silenced boolean NOT NULL DEFAULT false,
   status text NOT NULL,
   description text,
   lifecycle text,
   tooltip text,
+  status_reason text,
   icon text,
   type text NULL,
   owner text,
-  spec jsonb,
```
moshloop, line 44: Why are we removing this ?

pkg/db/migrations/3_topology.sql
```diff
@@ -2,55 +2,77 @@
...
+  topology_type text,
+  namespace text,
   labels jsonb null,
   hidden boolean NOT NULL DEFAULT false,
   silenced boolean NOT NULL DEFAULT false,
   status text NOT NULL,
   description text,
   lifecycle text,
   tooltip text,
+  status_reason text,
```
moshloop, line 63: It probably makes more sense to move this into summary

pkg/system_api.go
```diff
@@ -1,35 +1,91 @@
...
+	ID               uuid.UUID           `json:"id,omitempty" gorm:"default:generate_ulid()"` //nolint
+	Tooltip          string              `json:"tooltip,omitempty"`
+	Icon             string              `json:"icon,omitempty"`
+	Text             string              `json:"text,omitempty"`
+	Label            string              `json:"label,omitempty"`
+	Labels           types.JSONStringMap `json:"labels,omitempty"`
+	Owner            string              `json:"owner,omitempty"`
+	Components       Components          `json:"components,omitempty" gorm:"-"` //ignor  this
+	Properties       Properties          `json:"properties,omitempty" gorm:"type:properties"`
+	Summary          v1.Summary          `json:"summary,omitempty" gorm:"type:summary"` // ignor ethis
```
moshloop, line 37:
```suggestion
	Components       Components          `json:"components,omitempty" gorm:"-"` 
	Properties       Properties          `json:"properties,omitempty" gorm:"type:properties"`
	Summary          v1.Summary          `json:"summary,omitempty" gorm:"type:summary"` 
```

pkg/db/migrations/3_topology.sql
Kaitou786: I found the If not exists is beneficial for the local development envs... considering the case when I just need to alter one of the migration script; delete the goose_db_version and the target table... If I don't have the if not exists I would either need to do delete all the tables or use a different DB altogether.

pkg/db/migrations/3_topology.sql
Kaitou786: Since we don't have a spec in the components?

Issue #470, opened 2022-04-28
Operator crashed after canary was deleted


BrendanGalloway:
After deleting a canary object that was monitoring an object that no longer existed, the oeprator started crashing with the following log:
```
2022-04-28T09:09:57.457Z	INFO	deleting canary sandbox/admin-console-sqluser
E0428 09:09:57.459479       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 1970 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x59d6ec0, 0xa115b10})
	/go/pkg/mod/k8s.io/apimachinery@v0.20.6/pkg/util/runtime/runtime.go:74 +0x85
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0xc017666540})
	/go/pkg/mod/k8s.io/apimachinery@v0.20.6/pkg/util/runtime/runtime.go:48 +0x75
panic({0x59d6ec0, 0xa115b10})
	/usr/local/go/src/runtime/panic.go:1038 +0x215
github.com/flanksource/canary-checker/pkg/db.DeleteCanary({{{0x541a552, 0x6}, {0xc01a376c20, 0x1b}}, {{0xc017c76348, 0x15}, {0x0, 0x0}, {0xc017c6f820, 0x7}, ...}, ...})
	/app/pkg/db/canary.go:82 +0x235
github.com/flanksource/canary-checker/pkg/controllers.(*CanaryReconciler).Reconcile(0xc0178c0f00, {0x7c13ff8, 0xc01a40bb00}, {{{0xc017c6f820, 0x7}, {0xc017c76348, 0x15}}})
	/app/pkg/controllers/canary_controller.go:86 +0x634
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc0178c0fa0, {0x7c13f50, 0xc01594e040}, {0x5b70780, 0xc017666540})
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.7.2/pkg/internal/controller/controller.go:263 +0x303
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc0178c0fa0, {0x7c13f50, 0xc01594e040})
[...]

PR #471, opened 2022-04-28
fix: improve safety around db deletion

BrendanGalloway:
#470

### DIFF
--- a/pkg/db/canary.go
+++ b/pkg/db/canary.go
@@ -77,9 +77,13 @@ func PersistCheck(check pkg.Check) (string, error) {
 
 func DeleteCanary(canary v1.Canary) error {
 	logger.Infof("deleting canary %s/%s", canary.Namespace, canary.Name)
-	model := pkg.CanaryFromV1(canary)
+	model, err := pkg.CanaryFromV1(canary)
+	if err != nil {
+		return err
+	}
 	deleteTime := time.Now().Time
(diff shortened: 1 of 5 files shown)


pkg/db/canary.go
```diff
@@ -77,9 +77,13 @@ func PersistCheck(check pkg.Check) (string, error) {
...
 func DeleteCanary(canary v1.Canary) error {
 	logger.Infof("deleting canary %s/%s", canary.Namespace, canary.Name)
-	model := pkg.CanaryFromV1(canary)
+	model, err := pkg.CanaryFromV1(canary)
+	if err != nil {
+		return err
+	}
 	deleteTime := time.Now().Time
-	tx := Gorm.Find(&model).Where("id = ?", *canary.Status.PersistedID).UpdateColumn("deleted_at", deleteTime)
+	persistedID := canary.GetPersistedID()
```
BrendanGalloway, line 85: Do I also need to check and error if this is "" ?

moshloop: If persistedID is nil or "" then we can skip the db delete

Kaitou786 (approved):

pkg/db/canary.go
moshloop, line 85:
```suggestion
	persistedID := canary.GetPersistedID()
	if persistedID == "" {
		return nil
	}
```

PR #481, opened 2022-05-03
feat: add elasticsearch checker. Fixes #383

### LINKED ISSUE #383: Elasticsearch check

yashmehrotra:
Proposed spec:
```yaml
kind: Canary
metadata:
  name: elasticsearch
spec:
  interval: 30
  elasticsearch:
    - url: https://elasticsearch-host.example.com:9200
      description: Elasticsearch checker
      index: myIndex
      query: |
        {
          "query": {
            "term": {
              "system.role": "api"
            }
          }
        }
      results: 4
      auth:
        username:
          value: admin
        password:
          value: secret
```

The checker will fire the query on the given index and compare the count of results.

Remaining things to do in this PR:
- [x] Generate CRDs
- [x] Test entire flow
- [x] Update documentation
- [x] Linting


fixtures/datasources/elasticsearch_pass.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: canaries.flanksource.com/v1
+kind: Canary
+metadata:
+  name: elasticsearch
+spec:
+  interval: 30
+  elasticsearch:
+    - url: https://elasticsearch-host.example.com:9200
```
moshloop, line 8:
```suggestion
    - url: http://elasticsearch.default.svc:9200
```

checks/elasticsearch.go
```diff
@@ -0,0 +1,96 @@
...
+		//es.Search.WithContext(ctx),
+		es.Search.WithIndex(check.Index),
+		es.Search.WithBody(body),
+	)
+
+	if err != nil {
+		return results.ErrorMessage(err)
+	}
+
+	defer res.Body.Close()
```
Kaitou786, line 63: I would add this defer before the error check. Since in case of error we won't be closing the body since it is returning before adding this defer statement

Kaitou786 (commented):
Will review it more. But from a glance it looks good.

Kaitou786:
Also currently we don't have the conventional commit check enabled
But as general rule of thumb we would need all the commits to be conventional commits as part of the PR.
Since we don't squash before merge all the commits remains as it is in the main commit history

checks/elasticsearch.go
yashmehrotra: Makes sense, good catch :)

yashmehrotra:
> Also currently we don't have the conventional commit check enabled
But as general rule of thumb we would need all the commits to be conventional commits as part of the PR.
Since we don't squash before merge all the commits remains as it is in the main commit history

@Kaitou786 Once the PR is approved, I'll squash/rewrite all the commits to adhere to the convention

api/v1/checks.go
```diff
@@ -382,6 +382,32 @@ type Mongo struct {
...
+[include:datasources/elasticsearch_pass.yaml]
+*/
+
+type Elasticsearch struct {
+	ElasticsearchCheck `yaml:",inline" json:",inline"`
+}
+
+type ElasticsearchCheck struct {
+	Description `yaml:",inline" json:",inline"`
+	URL         string `yaml:"url" json:"url,omitempty" template:"true"`
```
Kaitou786, line 395: I would think this is a required field.. so for making it so in the CRD, remove the `omitempty` from the notation... also the formatting seems a bit wrong... can run `go fmt ./...`

Kaitou786:
@yashmehrotra  can you also check the tests... seems a runtime error on the new elasticsearch check... have you tested it locally?

yashmehrotra:
@Kaitou786 This is still WIP. Tests, formatting, CRD gen and documentation is remaining. I'll let you know when the PR is ready for full reiview.

Kaitou786:
> @Kaitou786 This is still WIP. Tests, formatting, CRD gen and documentation is remaining. I'll let you know when the PR is ready for full reiview.

Ah yeah, just saw that as remaining items on the description :)

api/v1/checks.go
```diff
@@ -382,6 +382,32 @@ type Mongo struct {
...
+
```
Kaitou786, line 396: Can we also add: `Templatable `yaml:",inline" json:",inline"`` in the struct. This gives more flexibilty in terms of tests with expr and JS... also can display it nicely on the dashboard

checks/elasticsearch.go
```diff
@@ -0,0 +1,93 @@
...
+				res.Status(),
+				e["error"].(map[string]interface{})["type"],
+				e["error"].(map[string]interface{})["reason"],
+			))
+		}
+	}
+
+	// We are closing the body after error as the Body object is not set in case of error
+	// leading to nil pointer errors
+	defer res.Body.Close()
```
Kaitou786, line 77: I would add this just after we create the res... since in case of res.IsError() we would be returning with a close body.

yashmehrotra: As mentioned in the comment, if we put this before res.IsError and there is an error, we will get a nil pointer dereference error ... previously it was being used when the res object was being created but it was giving runtime errors in case of a failure scenario

api/v1/checks.go
yashmehrotra: Sure

checks/elasticsearch.go
Kaitou786: Ah I see

PR #497, opened 2022-05-26
feat: add componenet selector

Kaitou786:


pkg/system_api.go
```diff
@@ -13,13 +12,20 @@ import (
...
 	"gorm.io/gorm"
 	"gorm.io/gorm/clause"
 	"gorm.io/gorm/schema"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+var json = jsontime.ConfigWithCustomTimeFormat
+
+func init() {
+	jsontime.AddTimeFormatAlias("postgres_timestamp", "2006-01-02T15:04:05.999999999")
```
moshloop, line 26:
```suggestion
	jsontime.AddTimeFormatAlias("postgres_timestamp", "2006-01-02T15:04:05.999")
```

pkg/topology/query.go
```diff
@@ -2,18 +2,25 @@ package topology
...
 	"github.com/flanksource/canary-checker/pkg"
 	"github.com/flanksource/canary-checker/pkg/db"
 	"github.com/flanksource/commons/logger"
-	"github.com/pkg/errors"
 )
 
+var json = jsontime.ConfigWithCustomTimeFormat
+
+func init() {
+	jsontime.AddTimeFormatAlias("postgres_timestamp", "2006-01-02T15:04:05.999999999")
```
moshloop, line 21: Isn't this duplicated ?

Kaitou786: They are initialized in two different packages... one is the `topology` package and another one is the `pkg` package

PR #531, opened 2022-06-20
fix: migrate ingress to use networking/v1 api

paddatrapper:


paddatrapper:
Rate limit on the CI restic pull. Will try run them again in 10 minutes

paddatrapper:
@moshloop the parts pertaining to the ingress update are now passing

moshloop:
@paddatrapper I think lets rather just hardcode to Prefix, these checks are going to be replaced anyway so I want to avoid too many CRD changes

PR #547, opened 2022-07-04
feat: remove the systems from the heirarchy

Kaitou786:


pkg/db/migrations/3_topology.sql
```diff
@@ -109,5 +83,5 @@ language plpgsql;
 DROP TABLE component_relationships;
 DROP FUNCTION lookup_component_by_property;
 DROP TABLE components;
-DROP TABLE systems;
+-- DROP TABLE systems;
```
moshloop, line 86:
```suggestion
```

pkg/db/topology.go
```diff
@@ -193,28 +181,14 @@ func DeleteSystemTemplate(systemTemplate *v1.SystemTemplate) error {
...
-	tx := Gorm.Find(systemModel).Where("system_template_id = ?", systemTemplateID).UpdateColumn("deleted_at", deleteTime)
-	if tx.Error != nil {
-		return "", tx.Error
-	}
-	return systemModel.ID.String(), nil
+	return DeleteComponnents(systemTemplate.GetPersistedID(), deleteTime)
 }
 
 // DeleteComponents deletes all components associated with a system
 func DeleteComponnents(systemID string, deleteTime time.Time) error {
```
moshloop, line 188: A system template might generate many components - we would want to delete by path
```suggestion
func DeleteComponents(componentId string, deleteTime time.Time) error {
```

Kaitou786: @moshloop 

This function is called when a systemTemplate object is deleted... In that case, all the components need to be deleted that were linked to that systemTemplateID... For the components that stops existing....i.e. a pod got  deleted in the cluster... it is handled in the difference as part of the sync job.

pkg/controllers/system_jobs.go
```diff
@@ -34,27 +34,31 @@ func (job SystemJob) Run() {
...
-		system.SystemTemplateID = &systemTemplateID
-		systemID, comps, err := db.PersistSystem(system)
-		if err != nil {
-			logger.Errorf("error persisting the system: %v", err)
-		}
-		dbComps, err := db.GetActiveComponentsWithSystemID(systemID)
+	components := topology.Run(opts, job.SystemTemplate)
+	systemTemplateID, err := uuid.Parse(job.SystemTemplate.GetPersistedID())
+	if err != nil {
+		logger.Errorf("error finding the systemTemplateID")
```
moshloop, line 40: When wouldn't we have a persistent id ?

Kaitou786: I don't see a scenario where this would execute but in case there is some scenario we would want to log it? and handle it accordingly?

The other alternative would be to use uuid.MustParse which would panic in case of an error.... or to just through the error variable all together

pkg/controllers/system_jobs.go
```diff
@@ -34,27 +34,31 @@ func (job SystemJob) Run() {
...
+	for _, component := range components {
+		component.Name = job.SystemTemplate.Name
+		component.Namespace = job.SystemTemplate.Namespace
+		component.Labels = job.SystemTemplate.Labels
+		component.SystemTemplateID = &systemTemplateID
+		componentsIDs, err := db.PersistComponent(component)
 		if err != nil {
-			logger.Errorf("error getting components for system: %v", err)
-			continue
+			logger.Errorf("error persisting the component: %v", err)
```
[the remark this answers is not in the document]
Kaitou786: Not sure how we are ignoring it...it wouldn't matter even if I were to add `continue` here... since this is the last thing happening in the loop?

PR #571, opened 2022-07-26
feat: add summary trigger; update status and summary on the DB

Kaitou786:


pkg/system_api.go
```diff
@@ -493,6 +508,20 @@ func (components Components) Summarize() v1.Summary {
 	return s
 }
 
+func (component Component) GetStatus() string {
+	if component.Summary.Healthy > 0 && component.Summary.Unhealthy > 0 {
+		return "warning"
```
yashmehrotra, line 513: Should we make these as constants ? Saw this being used at other places too

Kaitou786: Yeah, I think so too... we should be creating a status type and have it only have these values

pkg/db/topology.go
```diff
@@ -169,6 +169,18 @@ func PersistComponent(component *pkg.Component) ([]uuid.UUID, error) {
 	return persistedComponents, tx.Error
 }
 
+func UpdateStatusAndSummarForComponent(id uuid.UUID, status string, summary v1.Summary) error {
+	component := &pkg.Component{}
+	tx := Gorm.Where("id = ?", id).Find(component)
```
moshloop, line 174: I think we can merge these 2 queries into a single one. 

ie.. `UPDATE component set status = :status, summary = :summary WHERE id = :id and status != :status and summary != summary`

This saves a round trip to the database, and 2 marshal/unmarshall/serialize steps. 
The return value of the update statement (no of rows) can then be used as a metric to track summary updates per minute.

pkg/topology/run.go
```diff
@@ -288,3 +288,25 @@ func ComponentRun() {
...
+	})
+	if err != nil {
+		logger.Errorf("error getting components: %v", err)
+		return
+	}
+	for _, component := range components.Walk() {
+		if component.Name == "cluster" {
+			fmt.Println("cluster is found")
+			fmt.Println(component.Summary)
+		}
```
moshloop, line 305:
```suggestion
```

PR #576, opened 2022-07-27
chore: always return component with matching id

Kaitou786:


pkg/system_api.go
```diff
@@ -109,7 +109,7 @@ func (components *Components) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
-func (components Components) CreateTreeStrcuture() Components {
+func (components Components) CreateTreeStrcuture(queryID, status string) Components {
```
moshloop, line 112: I think just pass through the entire params object, also there is a typo

Kaitou786: We can't pass the entire object at that creates the import cycle

moshloop: Ok - can we fix the typo ? CreateTreeStructure

Kaitou786: Pushed

PR #578, opened 2022-07-31
Canary selector

Kaitou786:


api/v1/db_types.go
```diff
@@ -12,8 +12,20 @@ import (
 	"gorm.io/gorm/schema"
 )
 
+const (
+	SQLServerType = "sqlserver"
+	PostgresType  = "postgres"
+	SqliteType    = "sqlite"
```
yashmehrotra, line 18: These are capital because we are exporting them too ?

fixtures/topology/canary-selector.yaml
```diff
@@ -0,0 +1,32 @@
...
+  labels:
+    canary: canary-selector
+spec:
+  type: Website
+  icon: Application
+  schedule: "@every 5m"
+
+  components:
+    - checks:
+       -  selector:
```
yashmehrotra, line 14: (nitpick) whitespace

pkg/cache/postgres.go
```diff
@@ -35,12 +36,12 @@ func (c *postgresCache) AddCheckStatus(check pkg.Check, status pkg.CheckStatus)
 	if err != nil {
 		logger.Errorf("error marshalling details: %v", err)
 	}
-
-	row := c.QueryRow(context.TODO(), "UPDATE checks SET last_runtime = NOW() WHERE canary_id = $1 AND type = $2 AND name = $3 RETURNING id", check.CanaryID, check.Type, check.GetName())
+	row := c.QueryRow(context.TODO(), "UPDATE checks SET last_runtime = NOW(), status = $1 WHERE canary_id = $2 AND type = $3 AND name = $4 RETURNING id", status.Status, check.CanaryID, check.Type, check.GetName())
```
yashmehrotra, line 39: Lets break it into multiple lines for readability ?

pkg/db/topology.go
```diff
@@ -136,6 +171,23 @@ func PersistComponentRelationship(relationship *pkg.ComponentRelationship) error
 	return tx.Error
 }
 
+func PersisteCheckComponentRelationships(relationships []*pkg.CheckComponentRelationship) error {
```
yashmehrotra, line 174: typo

pkg/db/topology.go
```diff
@@ -119,6 +140,20 @@ func GetSelectorID(selectors v1.ResourceSelectors) string {
...
+	}
+	hash := md5.Sum(data)
+	if err != nil {
+		logger.Errorf("Error hashing component canaries %v", err)
+		return ""
+	}
+	return hex.EncodeToString(hash[:])
+}
+
 func PersisComponentRelationships(relationships []*pkg.ComponentRelationship) error {
```
yashmehrotra, line 157: typo

api/v1/db_types.go
Kaitou786: There are other places where they might be used... for example the pkg.Properties is in a different package but we need those values there

pkg/db/migrations/2_checks.sql
```diff
@@ -22,11 +22,13 @@ CREATE TABLE checks(
 	description TEXT,
 	icon TEXT,
 	spec jsonb  NULL,
+	labels jsonb NULL,
 	owner text,
 	severity TEXT,
 	last_runtime TIMESTAMP,
 	next_runtime TIMESTAMP,
 	silenced_at TIMESTAMP NULL,
+	status boolean, -- status of last check executed
```
moshloop, line 31: Can we change this to a text as we could have different status e.g. warning, info etc..

pkg/db/migrations/9_canary_component_relationships.sql
```diff
@@ -0,0 +1,22 @@
+-- +goose Up
+-- +goose StatementBegin
+
+CREATE TABLE check_component_relationships(
```
moshloop, line 4: We need to add this table into the seed, and all the other changes
and then in the seed we can insert into canary_checker_version 4, so that it skips the migrations for new db's

PR #607, opened 2022-09-08
fix: deletion logic for checks in case a subset of check is removed f…

Kaitou786:
…rom a single canary

### DIFF
--- a/pkg/controllers/canary_controller.go
+++ b/pkg/controllers/canary_controller.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/flanksource/canary-checker/pkg/db"
+	"github.com/google/uuid"
 
 	"github.com/flanksource/canary-checker/api/context"
 	v1 "github.com/flanksource/canary-checker/api/v1"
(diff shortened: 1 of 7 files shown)


pkg/controllers/canary_controller.go
```diff
@@ -225,3 +243,27 @@ func (r *CanaryReconciler) includeNamespace(namespace string) bool {
 	}
 	return false
 }
+
+func getCheckIDsForCanary(canaryID uuid.UUID) []string {
+	checks, _ := db.GetAllChecksForCanary(canaryID)
```
yashmehrotra, line 248: no error handling required ?

pkg/controllers/canary_controller.go
```diff
@@ -225,3 +243,27 @@ func (r *CanaryReconciler) includeNamespace(namespace string) bool {
...
+	for _, x := range b {
+		mb[x] = struct{}{}
+	}
+	var diff []string
+	for _, x := range a {
+		if _, found := mb[x]; !found {
+			diff = append(diff, x)
+		}
+	}
+	return diff
```
yashmehrotra, line 268: We can rename it to `SetDifference` and treat it like a utility function

yashmehrotra: No need for giving it the checks and status context

pkg/controllers/canary_controller.go
Kaitou786: Even in case, the error happens, we won't be stopping the reconcile loop or doing anything here. So I thought would be better to just ignore it

pkg/controllers/canary_controller.go
Kaitou786: yeah, make sense. moved io to utils

PR #619, opened 2022-09-21
feat: add sftp check

Kaitou786:


checks/folder_smb.go
```diff
@@ -141,6 +113,7 @@ func getServerDetails(serverPath string) (server, sharename, searchPath string,
 	case 2:
 		sharename = serverDetails[1]
 		searchPath = "."
+		fmt.Println("returning from here")
```
yashmehrotra, line 116: I think you missed removing this

fixtures/datasources/SFTP/sftp_pass.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: canaries.flanksource.com/v1
+kind: Canary
+metadata:
+  name: sftp-pass
+spec:
+  interval: 30
+  folder:
+    - path: /Users/tarun/flanksource/canary-checker
```
yashmehrotra, line 8: you can use `/tmp` for a generic path that exists everywhere

checks/folder_sftp.go
```diff
@@ -0,0 +1,60 @@
...
+
+func sshConnect(host string, port int, user string, password string) (*ssh.Client, error) {
+	config := &ssh.ClientConfig{
+		User: user,
+		Auth: []ssh.AuthMethod{
+			ssh.Password(password),
+		},
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+	}
+	client, err := ssh.Dial("tcp", host+":"+strconv.Itoa(port), config)
```
yashmehrotra, line 55: (nitpick: can-ignore):
`client, err := ssh.Dial("tcp", fmt.Sprintf("%s:%d", host, port), config)`

checks/folder_sftp.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return client, nil
```
yashmehrotra, line 59: Can just `return ssh.Dial(...)`

checks/folder_smb.go
Kaitou786: yeah xD

fixtures/datasources/SFTP/sftp_pass.yaml
Kaitou786: Yeah updated... just added for a quick check

checks/folder_sftp.go
Kaitou786: Updated

checks/folder_sftp.go
Kaitou786: updated

checks/folder_check.go
```diff
@@ -9,17 +9,21 @@ import (
...
-	Oldest  os.FileInfo
-	Newest  os.FileInfo
-	MinSize os.FileInfo
-	MaxSize os.FileInfo
-	Files   []os.FileInfo
+	Oldest                os.FileInfo
+	Newest                os.FileInfo
+	MinSize               os.FileInfo
+	MaxSize               os.FileInfo
+	SupportsTotalSize     bool
```
moshloop, line 22: This would part of the interface definition, each implementation would return a value on whether they support it or not

Kaitou786: We don't have a standard interface currently used by all the folder checks?

We can make a function on the Connection though?

PR #628, opened 2022-09-28
fix: change topology query and add nil check handling for configs

yashmehrotra:


pkg/db/config_db.go
```diff
@@ -85,6 +85,9 @@ func UpsertComponentConfigRelationship(componentID uuid.UUID, configs *pkg.Confi
 	if err != nil {
 		return err
 	}
+	if configs == nil {
+		return nil
+	}
```
Kaitou786, line 90: Should we move it as the first thing in the function. as in case the provided configs are nil we are unnecessarily making DB queries to get a configRelationship?

PR #636, opened 2022-10-07
fix: component config relationship

Kaitou786:


pkg/api.go
```diff
@@ -297,7 +298,7 @@ type Config struct {
 	Name         string         `json:"name,omitempty"`
 	Namespace    string         `json:"namespace,omitempty"`
 	Spec         *types.JSONMap `json:"spec,omitempty" gorm:"column:config"`
-	ExternalID   []string       `json:"external_id,omitempty"`
+	ExternalID   pq.StringArray `json:"external_id,omitempty" gorm:"type:text[]"`
```
moshloop, line 301: can we see if []string will work ?

Kaitou786: No, doesn't seem to work, get the same error:

```
name "external_id": unsupported Scan, storing driver.Value type string into type *[]string
```

PR #655, opened 2022-10-14
feat: add isLeaf flag for the leaf nodes and fix updates logic for co…

Kaitou786:
…mponents

### DIFF
--- a/pkg/db/migrations/103_is_leaf_components.sql
+++ b/pkg/db/migrations/103_is_leaf_components.sql
@@ -1,0 +1,3 @@
+-- +goose Up
+
+ALTER TABLE components ADD COLUMN is_leaf BOOL DEFAULT false;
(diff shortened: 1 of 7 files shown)


pkg/db/migrations/103_is_leaf_components.sql
```diff
@@ -0,0 +1,3 @@
+-- +goose Up
+
+ALTER TABLE components ADD COLUMN is_leaf BOOL DEFAULT false;
```
moshloop, line 3: Doesn't this need to be added to seed.sql and have the 99 script updated ?

moshloop (changes requested):

pkg/db/migrations/103_is_leaf_components.sql
Kaitou786: Yeah missed that.
Updated!

Issue #525, opened 2022-06-15
Topology - Open / Recent Incidents


moshloop:
On topology card show recent incidents.
If it makes senses, split to these tasks.
- Only for specific topology
- For the components in the topology.

moshloop:
As part of the seed scripts we can create a mock view of the incidents_by_component, and then update that view in incident_commander db always migration scripts. 

We could also move the topology query to the incident commander, and leave just a simple query in canary-checker with incidents, configs, etc..

Returning incidents (id, type, severity, name) should be optional (includeIncidents) with the default on.

PR #656, opened 2022-10-15
feat: add incidents in topology query

yashmehrotra:
Fixes: #525

### DIFF
--- a/pkg/topology/query.go
+++ b/pkg/topology/query.go
@@ -350,7 +350,6 @@ func getConfigForComponents() string {
             'name', config_items.name,
             'external_type', config_items.external_type,
             'config_type', config_items.config_type,
-            'config', config_items.config,
             'cost_per_minute', config_items.cost_per_minute,
             'cost_total_1d', config_items.cost_total_1d,
             'cost_total_7d', config_items.cost_total_7d,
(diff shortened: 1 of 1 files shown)


pkg/topology/query.go
```diff
@@ -331,16 +335,29 @@ func filterComponentsWithDepth(components []*pkg.Component, depth int) []*pkg.Co
...
+        WHERE check_component_relationships.component_id = components.id AND check_component_relationships.deleted_at IS NULL
+        GROUP BY check_component_relationships.component_id
+    ) :: jsonb`
 }
 
 func getConfigForComponents() string {
-	return `
-       (SELECT json_agg(config_items) from config_items
+	return `(
+        SELECT json_agg(config_items) from config_items
```
moshloop, line 348: We shouldn't be using the full config_items table it will return too much data

yashmehrotra: Lets return: `id, name, external_type, config_type, config, cost_per_minute, cost_total_1d, cost_total_7d, cost_total_30d`

pkg/topology/query.go
```diff
@@ -331,16 +335,45 @@ func filterComponentsWithDepth(components []*pkg.Component, depth int) []*pkg.Co
...
 func getConfigForComponents() string {
-	return `
-       (SELECT json_agg(config_items) from config_items
+	return `(
+        SELECT json_agg(json_build_object(
+            'id', config_items.id,
+            'name', config_items.name,
+            'external_type', config_items.external_type,
+            'config_type', config_items.config_type,
+            'config', config_items.config,
```
moshloop, line 353: We want to specifically exclude this as it can very large

PR #658, opened 2022-10-17
Component tests

Kaitou786:


moshloop:
@Kaitou786  I think we should split the removal of the UI into a speerate PR, - it looks like the test script needs a `set -e`

.github/workflows/topology_tests.yml
```diff
@@ -0,0 +1,37 @@
...
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }} # only needed if version is 'latest'
+        id: install
+      - name: Install Karina
+        run: make .bin/karina
+      - name: Build and Cache Image
+        run: |
+          # pull image if exists
+          echo ${{ secrets.GITHUB_TOKEN }} | docker login ghcr.io -u $GITHUB_ACTOR --password-stdin
+          docker pull ghcr.io/$GITHUB_REPOSITORY/canary-checker:ci || true
```
moshloop, line 26:
```suggestion
          docker pull docker.io/flanksource/canary-checker:ci || true
```

Kaitou786: This is image is meant for cache

Should we be putting it on dockerhub?

moshloop: Yes, we dont have gcp keys to push to ghcr

moshloop: n.m I though it was gcr

PR #695, opened 2022-11-30
feat: support javascript for config lookup

### LINKED ISSUE #687: Config lookup by expression
Support arbitrary Templates for property config lookups, 

e.g.  instead of

```yaml
     - name: Version
            headline: true
            configLookup:
              config:
                type: File
                name: Configuration.properties
                labels:
                  namespace: ${.properties.namespace}
              field: $["application.buildversion"]
```

```yaml
   -  name: Version
      template:
          javascript: configs['File/Configuration.properties'].config['application.buildversion']
```

Fields that should be available

* self
* self.properties
* configs ( A map of all related configs with `type/name` key )
*

yashmehrotra:
Fixes #687


fixtures/topology/k8s-system.yaml
```diff
@@ -36,6 +36,10 @@ spec:
                 name: nodemetrics
                 display:
                   javascript: JSON.stringify(k8s.getNodeMetrics(results))
+        - name: cluster-name
+          configLookup:
+            display:
+              javascript: config.getConfigItem("EKS/flanksource-canary-cluster").properties["name"]
```
moshloop, line 42: GetConfigItem by type/name is useful, but the primary use case would be to get the config item related to the component

templating/template.go
```diff
@@ -42,6 +44,31 @@ func Template(environment map[string]interface{}, template v1.Template) (string,
...
+			}
+			configItem, err := db.FindConfig(configItemParams)
+			if err != nil {
+				logger.Errorf("Error fetching config item for js: %v", err)
+				emptyObj, _ := vm.ToValue(map[string]string{})
+				return emptyObj
+			}
+
+			configObject := map[string]interface{}{
+				"configs":    *configItem.Spec,
```
moshloop, line 62: Why can't we just return the configItem  directly ?

fixtures/topology/k8s-system.yaml
yashmehrotra: configItemByType("EKS") -> [array ConfigItems]
configGetItem()

templating/template.go
yashmehrotra: @moshloop In the issue description, you mentioned we should expose keys `properties` and `configs`

fixtures/topology/k8s-system.yaml
```diff
@@ -36,6 +36,15 @@ spec:
                 name: nodemetrics
                 display:
                   javascript: JSON.stringify(k8s.getNodeMetrics(results))
+        - name: cluster-name-1
+          configLookup:
+            display:
+              javascript: configGetItem("EKS", "flanksource-canary-cluster").properties["name"]
```
yashmehrotra, line 42: getConfigItem

fixtures/topology/k8s-system.yaml
```diff
@@ -36,6 +36,15 @@ spec:
...
+        - name: cluster-name-2
+          configLookup:
+            display:
+              javascript: configGetItemsByTypeForComponent(componentID, "EKS")[0]["name"]
```
yashmehrotra, line 46: getConfigItems("EKS")

pkg/topology/run.go
```diff
@@ -202,6 +202,20 @@ func lookupConfig(ctx *ComponentContext, property *v1.Property, sisterProperties
 	prop := pkg.NewProperty(*property)
 
 	logger.Infof("Looking up config for %s => %s", property.Name, property.ConfigLookup.Config)
+
+	if property.ConfigLookup.Config == nil {
+		templateEnv := make(map[string]interface{})
+		if ctx.CurrentComponent != nil {
```
yashmehrotra, line 208: don't send comp id

fixtures/topology/k8s-system.yaml
```diff
@@ -36,6 +36,14 @@ spec:
                 name: nodemetrics
                 display:
                   javascript: JSON.stringify(k8s.getNodeMetrics(results))
+        - name: cluster-name-1
```
moshloop, line 39: Can we add a fixture for getConfigItems

PR #709, opened 2022-12-30
feat: sync costs for component

yashmehrotra:


pkg/db/topology.go
```diff
@@ -64,6 +65,53 @@ func GetAllComponentWithSelectors() (components pkg.Components, err error) {
...
+		ComponentID string
+		ConfigIDs   []string
+	}
+
+	var rows []result
+	err := Gorm.Raw(`
+        SELECT components.id as component_id, ARRAY_AGG(ccr.config_id) as config_ids
+        FROM components
+        INNER JOIN config_component_relationships ccr ON components.id = ccr.component_id
+        WHERE components.deleted_at IS NULL
```
moshloop, line 82: We still need to include deleted components from a cost perspective e.g. In an EKS cluster nodes can be deleted regularly but we still want to include them in the cost.

PR #719, opened 2023-01-06
Install instructions

BrendanGalloway:
@moshloop I have not updated the flanksource-ui.backend field yet.  We discussed jsut hardcoding the service name, so that the samevalue can be in the values file as default.  Everywhere else, things are named after the release name set as at runtime, so it  will be a big discrepancy if just that one has a hardcoded name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ To specify a username and password for the chart-managed Postgres server, create
 
 #### External Postgres Server
 
-In order to connect to an existing Postgres server, a database must be created on the server, along with a user that has <Not sure exactly what needs to be here> permissions
+In order to connect to an existing Postgres server, a database must be created on the server, along with a user that has administrator permissions for the database.git 
 
 |                     |                   |
 |---------------------|-------------------|
--- a/README.md
+++ b/README.md
@@ -109,6 +109,8 @@ Canary Checker should optimally be run connected to a dedicated Postgres Server,
 
 ##### Embedded database (default):
 
+|                     |                   |
+|---------------------|-------------------|
 | db.external.enabled | `false` (default) |
 | db.embedded.storageClass | Set to name of a storageclass available in the cluster |
 | db.embedded.storage | Set to volume of storage to request |
@@ -117,6 +119,8 @@ The Canary Checker statefulset will be configured to start an embedded postgres 
 
 ##### Fully automatic Postgres Server creation
 
+|                     |                   |
+|---------------------|-------------------|
 | db.external.enabled | `true` |
 | db.external.create  | `true` |
 | db.external.storageClass | Set to name of a storageclass available in the cluster |
@@ -130,6 +134,8 @@ To specify a username and password for the chart-managed Postgres server, create
 
 In order to connect to an existing Postgres server, a database must be created on the server, along with a user that has <Not sure exactly what needs to be here> permissions
 
+|                     |                   |
+|---------------------|-------------------|
 | db.external.enabled | `true` |
 | db.external.create  | `false` |
 | db.external.secretKeyRef.name | Set to name of name of secret that contains a key containging the postgres connection URI |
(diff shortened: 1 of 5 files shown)


chart/Chart.yaml
```diff
@@ -8,8 +8,8 @@ maintainers:
   - name: Flanksource
     url: https://www.flanksource.com
 icon: https://avatars.githubusercontent.com/u/58787470
-dependencies:
```
moshloop, line 11: Why do we comment this out ?

BrendanGalloway: THat was supposed to just be temporary while I templated the chart.  Have fixed

README.md
```diff
@@ -80,13 +80,110 @@ Canary Checker is a Kubernetes native multi-tenant synthetic monitoring system.
...
+To specify a username and password for the chart-managed Postgres server, create a secret in the namespace that the chart will install to, named `postgres-connection`, which contains `POSTGRES_USER` and `POSTGRES_PASSWORD` keys.
+
+#### Prexisting Server
+
+In order to connect to an existing Postgres server, a database must be created on the server, along with a user that has <Not sure exactly what needs to be here> permissions
+
+| db.external.enabled | `true` |
+| db.external.create  | `false` |
+| db.external.secretKeyRef.name | Set to name of name of secret that contains a key containging the postgres connection URI
+| db.external.secretKeyREf.key | Set to the name of the key in the secret that contains the postgres connection URI |
```
paddatrapper, line 136:
```suggestion
| db.external.secretKeyRef.key | Set to the name of the key in the secret that contains the postgres connection URI |
```

README.md
```diff
@@ -80,13 +80,110 @@ Canary Checker is a Kubernetes native multi-tenant synthetic monitoring system.
...
+### Add the Flanksource helm repository
+
+```bash
+helm repo add flanksource https://flanksource.github.io/charts
+helm repo update
+```
+
+### Configurable fields
+
+See the [values file](chart/values.yaml) for the full list of configurable fields.  Mandatory configuration values are for the configuration of the database, and it is recommended to also configure the UI ingress.
```
paddatrapper, line 104: You will also have to configure the `flanksource-ui.backendURL`, currently it is undefined and flanksource UI is used by default. It probably should have a default value pointing at the canary-checker service deployed by this chart

BrendanGalloway: The service deployed by this chart uses RELEASE_NAME, which is only known at runtime.  So it's not really possible to set a default, unless the service name is hardcoded.  In which case things like the deployment name should also be to avoid inconsistency?   Not sure how to handle that correctly, will add to the docs that the two names need to be aligned

BrendanGalloway:
As I note, I have got a placeholder describing what setup needs to be done for an external postgres database @moshloop is there any guidance in terms of what permissions the postgres account needs, or anything else that needs to be set up?  I think with the helm managed DB, we connect with the admin account, but for something like cloudSQL, we should recommend specific permissions.

README.md
moshloop: I think we should just hardcode both of them

README.md
```diff
@@ -80,13 +80,114 @@ Canary Checker is a Kubernetes native multi-tenant synthetic monitoring system.
...
+| db.external.enabled | `true` |
+| db.external.create  | `true` |
+| db.external.storageClass | Set to name of a storageclass available in the cluster |
+| db.external.storage | Set to volume of storage to request |
+
+The helm chart will create a postgres server statefulset, with a random password and default port, along with a canarychecker database hosted on the server.
+
+To specify a username and password for the chart-managed Postgres server, create a secret in the namespace that the chart will install to, named `postgres-connection`, which contains `POSTGRES_USER` and `POSTGRES_PASSWORD` keys.
+
+#### Prexisting Server
```
moshloop, line 129:
```suggestion
#### External Postgres Database
```

README.md
```diff
@@ -80,13 +80,114 @@ Canary Checker is a Kubernetes native multi-tenant synthetic monitoring system.
...
+
+See the [values file](chart/values.yaml) for the full list of configurable fields.  Mandatory configuration values are for the configuration of the database, and it is recommended to also configure the UI ingress.
+
+#### DB
+
+Canary Checker should optimally be run connected to a dedicated Postgres Server, but can run an embedded postgres instance for development and testing.
+
+##### Embedded database (default):
+
+| db.external.enabled | `false` (default) |
```
moshloop, line 112: Can you check the table formatting?

README.md
BrendanGalloway: I changed the helper function to use 'canary-checker' rather than the release name, so objects should be named consistently across the board.  It's still possible to override the name, and have documented that the backend connection would have to be manually aligned if that is the case.

PR #720, opened 2023-01-09
Feat/show deleted components

gjagnoor:
`component_names_all` query fetches all the components in the database and returns them to the frontend with the `deleted_at` field, in addition to `id, external_id, type, name, created_at, updated_at, icon, and parent_id`. The front end then based on the user request sees either all, deleted, or existing components.


gjagnoor:
@moshloop Done! Sorry, it took me some time to figure this out. First time squashing.

PR #729, opened 2023-01-11
feat: job history for sync jobs

yashmehrotra:


moshloop:
@yashmehrotra I think we move it to duty + error log - this is one of things where I think its ok to continue on error

yashmehrotra:
So, you mean to say something like, `jobHistory.Persist(db.Gorm)` and inside that have a log line ? The qualm I have about that is that it will take the logging option away from the user. Is that something we want to do ?

pkg/topology/checks/component_check.go
```diff
@@ -21,12 +22,16 @@ func ComponentCheckRun() {
 		return
 	}
 	for _, component := range components {
+		jobHistory := models.NewJobHistory("ComponentCheckRelationshipSync", "component", component.ID.String()).Start()
```
moshloop, line 25: I don't think this should be per component - it would create a huge number of jobs + at some point we may want to optimize this into a single sql call

pkg/topology/configs/component_config.go
```diff
@@ -23,10 +24,14 @@ func ComponentConfigRun() {
 	}
 
 	for _, component := range components {
+		jobHistory := models.NewJobHistory("ComponentConfigRelationshipSync", "component", component.ID.String()).Start()
```
moshloop, line 27: Same here

pkg/topology/sync.go
```diff
@@ -23,26 +24,28 @@ func ComponentRun() {
 	}
 
 	for _, component := range components {
+		jobHistory := models.NewJobHistory("ComponentRelationshipSync", "component", component.ID.String()).Start()
```
moshloop, line 27: This will create too many jobs

pkg/topology/sync.go
```diff
@@ -56,11 +59,15 @@ func ComponentStatusSummarySync() {
 		return
 	}
 	for _, component := range components.Walk() {
+		jobHistory := models.NewJobHistory("ComponentStatusSummarySync", "component", component.ID.String()).Start()
```
moshloop, line 62: and this

PR #737, opened 2023-01-18
feat(topology): add elasticsearch topology

paddatrapper:


fixtures/topology/elastic.yaml
```diff
@@ -0,0 +1,185 @@
...
+                    secretKeyRef:
+                      name: elasticsearch-elastic-user
+        - name: load
+    - name: index
+      type: ElasticSearchIndex
+      properties:
+        - name: name
+        - name: documents
+          http:
+            - endpoint: http://elasticsearch-http.default.svc:9200/_stats
```
moshloop, line 93: We can hit the endpoint once, and then using javascript return a list of properties

paddatrapper: How do I achieve that?

moshloop: Have a look at https://github.com/flanksource/canary-checker/blob/master/templating/js/k8s.js#L176

PR #752, opened 2023-01-24
fix: postgres secret creation logic

BrendanGalloway:
Realised I correctly catered for the case where secret values were specified apriori, but didn't add in the extra logic to keep the password static if not.

### DIFF
--- a/chart/templates/postgres-secret.yaml
+++ b/chart/templates/postgres-secret.yaml
@@ -1,4 +1,4 @@
-{{- if eq .Values.db.external.enabled true }}
+{{- if eq .Values.db.external.create true }}
 apiVersion: v1
 kind: Secret
 metadata:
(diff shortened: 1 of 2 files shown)


chart/templates/postgres-secret.yaml
```diff
@@ -9,9 +9,10 @@ metadata:
     "helm.sh/resource-policy": "keep"
 type: Opaque
 stringData:
-  {{- $secretObj := ( lookup "v1" "Secret" .Release.Namespace "postgres-connection" ).data }}
-  {{- $user := (( get $secretObj "POSTGRES_USER" ) | b64dec ) | default "postgres" }}
-  {{- $password := (( get $secretObj "POSTGRES_PASSWORD" ) | b64dec ) | default ( randAlphaNum 32 ) }}
+  {{- $secretInj := ( lookup "v1" "Secret" .Release.Namespace "postgres-connection" ).data }}
+  {{- $secretObj := ( lookup "v1" "Secret" .Release.Namespace .Values.db.external.secretKeyRef.name ).data }}
+  {{- $user := (( get $secretInj "POSTGRES_USER" ) | b64dec ) | default ( get $secretObj "POSTGRES_USER" ) | b64dec ) | default "postgres" }}
+  {{- $password := (( get $secretInj "POSTGRES_PASSWORD" ) | b64dec )| default ( get $secretObj "POSTGRES_PASSWORD" ) | b64dec ) | default ( randAlphaNum 32 ) }}
```
yashmehrotra, line 15: Another `(` between `default ( get $secretObj`

PR #766, opened 2023-02-02
docs: explain deployment with an ephemeral database

paddatrapper:


docs/operator.md
```diff
@@ -5,6 +5,138 @@ hide:
...
+
+Canary Checker can be deployed to a Kubernetes cluster via Helm.
+
+```bash
+helm repo add flanksource https://flanksource.github.io/charts
+helm repo update flanksource
+```
+
+Create a values file based on
+[values.yaml](https://github.com/flanksource/canary-checker/blob/master/chart/values.yaml).
```
moshloop, line 18: Can we just specify the non-default values and db settings

PR #774, opened 2023-02-07
fix: logic for handling deletion, updating status and readding CRDs

yashmehrotra:


pkg/db/topology.go
```diff
@@ -276,39 +281,38 @@ func DeleteSystemTemplate(systemTemplate *v1.SystemTemplate) error {
...
+func DeleteComponentsWithSystemTemplate(systemTemplateID string, deleteTime time.Time) error {
 	logger.Infof("Deleting all components associated with system: %s", systemTemplateID)
 	componentsModel := &[]pkg.Component{}
-	tx := Gorm.Where("system_template_id = ?", systemTemplateID).Find(componentsModel).UpdateColumn("deleted_at", deleteTime)
-	DeleteComponentRelationshipForComponents(componentsModel, deleteTime)
+	if err := Gorm.Where("system_template_id = ?", systemTemplateID).Find(componentsModel).UpdateColumn("deleted_at", deleteTime).Error; err != nil {
+		return err
+	}
 	for _, component := range *componentsModel {
+		if err := DeleteComponentChildren(component.ID.String(), deleteTime); err != nil {
```
moshloop, line 295: Can we not do a delete where path like 'parent.%' ?

It probably also makes sense to move the deletion logic to duty - especially so that we can add tests for it

yashmehrotra: Have updated to delete using `path like %component_id`

> It probably also makes sense to move the deletion logic to duty - especially so that we can add tests for it
Yes, I'll create a New PR where I'll move all the topology db logic to duty

PR #775, opened 2023-02-07
feat: move topology query to duty

yashmehrotra:
Need to update `go.mod` after https://github.com/flanksource/duty/pull/16 is merged


pkg/topology/query.go
```diff
@@ -156,100 +105,13 @@ func (p TopologyParams) getID() string {
...
-	}
-	if params.Owner != "" {
-		args["owner"] = params.Owner
-	}
-	if params.Labels != "" {
-		args["labels"] = params.getLabels()
-	}
-
-	logger.Tracef("Querying topology (%s) => %s", params, query)
+	query, args := duty.TopologyQuery(duty.TopologyOptions{
```
moshloop, line 109: I think we should move all the filtering and unmarshalling into duty as well,

yashmehrotra: I don't think we can do it right now without creating a cyclic dependency.

For duty to Marshal, CreateTreeStructure and Flatten it would need to have the components type which it will have to import from canary-checker.

We should eventually move it to duty but I would want to do that separately since the scope of that would be huge

moshloop: Can we not move the components type to duty, and then embed it into the canary-checker type ?

yashmehrotra: Have created an issue to track this as discussed: https://github.com/flanksource/canary-checker/issues/793

PR #786, opened 2023-02-10
feat: Custom renderer

adityathebe:
Addresses https://github.com/flanksource/incident-commander/issues/235


pkg/system_api.go
```diff
@@ -166,6 +166,17 @@ func (components Components) Walk() Components {
 	return comps
 }
 
+type RenderComponent struct {
```
adityathebe, line 169: I created a new struct here. Wasn't sure if I should be using the existing `Component`

moshloop:
@adityathebe this needs to go incident-commander, and the struct changes into duty, Nothing UI related should be in canary-checker, eventually all of the topology related code will move to incident commander, and only health checks will be performed by canary-checker.

pkg/topology/query.go
```diff
@@ -198,6 +198,33 @@ func (p TopologyParams) GetComponentRelationWhereClause() string {
 	return s
 }
 
+func QueryRenderComponents(ctx context.Context, systemTemplateID string) ([]pkg.Renderers, error) {
+	query := `select spec->'renderers' from templates where id = $1`
+	args := map[string]interface{}{
+		"id": systemTemplateID,
+	}
+
+	rows, err := db.QueryNamed(ctx, query, args)
```
yashmehrotra, line 207: Don't use `QueryNamed` function, prefer using Gorm or the underlying pool object to query

pkg/api/renderer.go
```diff
@@ -0,0 +1,120 @@
...
+const componentRegistry = {
+	{{range $k, $v := .}}"{{$k}}": {{$k}},
+	{{end}}
+};
+`
+
+func init() {
+	tpl, err := template.New("registry").Parse(jsComponentRegistryTpl)
+	if err != nil {
+		panic(fmt.Errorf("error parsing template jsComponentRegistryTpl. %w", err))
```
yashmehrotra, line 116: Use logger.Fatalf instead of panic

pkg/api/renderer.go
```diff
@@ -0,0 +1,120 @@
...
+	if isProp {
+		prefix = "property"
+	}
+
+	if c.Type != "" {
+		return fmt.Sprintf("%s_%s_%s", prefix, c.Type, c.Name)
+	}
+
+	return fmt.Sprintf("%s_%s", prefix, c.Name)
+}
```
yashmehrotra, line 91: This can be a method of pkg.RenderComponent,  so we can use it like `renderComponent.Key(isProp)`

PR #831, opened 2023-02-28
feat: use duty for complete topology generation

yashmehrotra:
Merge once https://github.com/flanksource/duty/pull/45 is merged after bumping go


pkg/topology/query.go
```diff
@@ -1,231 +1,65 @@
...
-
-func parseItems(items string) []string {
-	if strings.TrimSpace(items) == "" {
-		return nil
-	}
-	return strings.Split(items, ",")
-}
+const DefaultDepth = 3
 
 type TopologyParams struct {
```
moshloop, line 17: Can we move this to duty as well as we would want to switch the api call to incident-commander so that we can still return the topology even if canary checker is down

yashmehrotra: Have removed this struct, we now use duty.TopologyParams everywhere

Issue #826, opened 2023-02-24
Handle bad canary parsing and record event


yashmehrotra:
Currently canary checker will break at startup if the spec unmarshalling fails (To reproduce this, create the canary in db with spec: `{"interval": "300", "cloudwatch": [{"name": null}]`

This will break canary-checker since interval should be `int`.

Handle such errors gracefully and update job history for these

PR #833, opened 2023-02-28
fix: handle invalid canary spec and update job history

yashmehrotra:
Fixes: https://github.com/flanksource/canary-checker/issues/826

### DIFF
--- a/pkg/db/canary.go
+++ b/pkg/db/canary.go
@@ -43,7 +43,11 @@ func GetAllCanaries() ([]v1.Canary, error) {
 		return nil, err
 	}
 	for _, _canary := range _canaries {
-		canaries = append(canaries, *_canary.ToV1())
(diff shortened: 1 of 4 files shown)


pkg/db/canary.go
```diff
@@ -43,7 +43,11 @@ func GetAllCanaries() ([]v1.Canary, error) {
 		return nil, err
 	}
 	for _, _canary := range _canaries {
-		canaries = append(canaries, *_canary.ToV1())
+		c := _canary.ToV1()
+		if c == nil {
+			continue
```
yashmehrotra, line 48: @moshloop Should I just return an error here and track it to job history too, the only way this arises is when spec read fails ?

moshloop: We might need to maintain a error list and then return that together with the successful parses

PR #842, opened 2023-03-03
feat: api for check status summary

yashmehrotra:


pkg/cache/postgres.go
```diff
@@ -82,6 +85,76 @@ func (c *postgresCache) Query(q QueryParams) (pkg.Checks, error) {
 	return q.ExecuteSummary(db.Pool)
 }
 
+func (c *postgresCache) QuerySummary() (pkg.Checks, error) {
```
moshloop, line 88: Shouldn't this be in duty alongside the materialized refresh function?

PR #863, opened 2023-03-15
fix: skip templating if check fails

### LINKED ISSUE #374: ConnectionFailureMode
- Failed (don't run exprs) 
- Invalid (+ fire different event, don't run exprs)
   + don't include status in lastResult function 
   + check status become invalid
- Continue (run exprs to determine status)

yashmehrotra:
Fixes: https://github.com/flanksource/canary-checker/issues/374

### DIFF
--- a/checks/runchecks.go
+++ b/checks/runchecks.go
@@ -29,14 +29,21 @@ func RunChecks(ctx *context.Context) []*pkg.CheckResult {
 
 func transformResults(ctx *context.Context, in []*pkg.CheckResult) (out []*pkg.CheckResult) {
 	for _, r := range in {
+		// If check fails, do not apply transform
+		if !r.Pass {
+			out = append(out, r)
+			continue
+		}
+
 		transformed, err := transform(ctx, r)
 		if err != nil {
 			r.Failf("transformation failure: %v", err)
 			out = append(out, r)
-		} else {
-			for _, t := range transformed {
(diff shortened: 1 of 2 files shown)


checks/runchecks.go
```diff
@@ -29,14 +29,21 @@ func RunChecks(ctx *context.Context) []*pkg.CheckResult {
 
 func transformResults(ctx *context.Context, in []*pkg.CheckResult) (out []*pkg.CheckResult) {
 	for _, r := range in {
+		// If check fails, do not apply transform
+		if !r.Pass {
```
moshloop, line 33: !pass doesn't necessarily means it failed to connect - it could have failed a threshold, response code/body check etc.

yashmehrotra: I went through the checks code and all the errors are raised via `.Failf` or `.ErrorMessage` (which calls `.Failf` inside) all of which sets pass = false.

If not for r.Pass, what should we check against ?

moshloop: We can add another field called Error = true that ErrorMessage sets - however we might need to update some of the checks to use ErrorMessage instead of Failf for connection errors

moshloop (changes requested):
Still need to add the error handling

adityathebe:
https://github.com/flanksource/canary-checker/pull/3028

PR #898, opened 2023-03-27
fix: change type of child components

yashmehrotra:


api/v1/component_types.go
```diff
@@ -35,7 +34,9 @@ type ComponentSpec struct {
 	// forEach property to iterate over the results to further enrich each component.
 	Lookup *CanarySpec `json:"lookup,omitempty"`
 	// Create new child components
-	Components []json.RawMessage `json:"components,omitempty"`
+	// TODO: This should be a component spec but due to kubernetes crd size limitations
+	// we are keeping it as a string and unmarshalling it at runtime
+	Components string `json:"components,omitempty"`
```
moshloop, line 39: Why not json raw message ?

yashmehrotra: Because as per the openapi spec, `[]bytes` needs to be given as base64 encoded input.

https://swagger.io/docs/specification/data-models/data-types/#strings

```
String Formats
An optional format modifier serves as a hint at the contents and format of the string. OpenAPI defines the following built-in string formats:

    date – full-date notation as defined by [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6), for example, 2017-07-21
    date-time – the date-time notation as defined by [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6), for example, 2017-07-21T17:32:28Z
    password – a hint to UIs to mask the input
    byte – base64-encoded characters, for example, U3dhZ2dlciByb2Nrcw==
````

File: api/v1/component_types.go [hunk 2]
```diff
@@ -50,6 +51,9 @@ type ComponentSpec struct {
 	ForEach *ForEach `json:"forEach,omitempty"`
 }
 
+// +kubebuilder:validation:Type=object
+type ComponentSpecObject ComponentSpec
+
```
yashmehrotra, line 56: Have to declare a new type indirection because kubebuilder didn't process using the same type  `ComponentSpec`

Issue #424, opened 2022-03-11
Config-db query check


moshloop:

moshloop:
Move to using the duty API's

PR #899, opened 2023-03-27
Config db query check

adityathebe:
Closes https://github.com/flanksource/canary-checker/issues/424


checks/config_db.go
```diff
@@ -26,38 +21,25 @@ func (c *ConfigdbChecker) Run(ctx *canaryContext.Context) pkg.Results {
...
 	if err != nil {
-		return results.Failf("Failed to read the response body: %v", err)
-	}
-	queryResult := ConfigDBQueryResult{}
-	if err := json.Unmarshal(body, &queryResult); err != nil {
-		results.Failf("failed to unmarshal the response body into QueryResult: %v", err)
+		return results.Failf("failed running query: %v", err)
 	}
+
+	queryResult := ConfigDBQueryResult{Results: res}
```
adityathebe, line 42: For `ConfigDBQueryResult`, I haven't populated `Count` & `Columns` because the `duty.Query` only returns results. I can make that change to `Query` func if that's necessary.

PR #900, opened 2023-03-27
feat: Check statuses aggregation jobs

adityathebe:


api/v1/constants.go
```diff
@@ -1,13 +1,17 @@
...
+	ComponentStatusSummarySyncSchedule      = "@every 1m"
+	ComponentCheckSchedule                  = "@every 2m"
+	ComponentConfigSchedule                 = "@every 2m"
+	ComponentCostSchedule                   = "@every 1h"
+	CheckStatusSummarySchedule              = "@every 1m"
+	CheckStatusesAggregate1hSchedule        = "@every 1h"
+	CheckStatusesAggregate1dSchedule        = "@every 24h"
+	CheckStatuses1hAggregatedDeleteSchedule = "@every 24h"
+	CheckStatuses1dAggregatedDeleteSchedule = "@every 24h"
+	CheckStatusDeleteSchedule               = "@every 24h"
```
moshloop, line 16: I would merge these 3 into a single job and schedule

pkg/db/check_statuses.go
```diff
@@ -29,3 +41,160 @@ func DeleteOldCheckStatuses() {
...
+		SELECT
+		check_statuses.check_id,
+		date_trunc(?, "time"),
+		count(*) AS total_checks,
+		count(*) FILTER (WHERE check_statuses.status = TRUE) AS passed,
+		count(*) FILTER (WHERE check_statuses.status = FALSE) AS failed,
+		SUM(duration) AS duration
+	FROM check_statuses 
+	LEFT JOIN checks ON check_statuses.check_id = checks.id
+	WHERE checks.created_at > NOW() - INTERVAL '1 day' * ?
```
moshloop, line 146: We should only be querying for the time range of 1h or 1d, and inserting the results - We shouldn't be return all the 1h values for the last day and then try upsert 24 values.

api/v1/constants.go
adityathebe: That's how I did it initially but these are 3 separate elementary operations. I'm not sure how to mark the status of the job if they fail partially.

moshloop: You can have different job names, all with the same schedule

pkg/db/check_statuses.go
moshloop: But at the expense of doing 24x as much work and requiring a primary key - maybe just run it for the last 3h

api/v1/constants.go
adityathebe: 😁 that was quite silly of me.

PR #904, opened 2023-03-28
fix: lookup handling for child components

yashmehrotra:


pkg/topology/run.go
```diff
@@ -359,14 +361,8 @@ func Run(opts TopologyRunOptions, s v1.SystemTemplate) []*pkg.Component {
 				component.Components = append(component.Components, components...)
 				continue
 			}
-			group := pkg.NewComponent(comp)
-			group.Components = append(group.Components, components...)
-			if len(group.Components) > 0 {
-				group.Summary = group.Components.Summarize()
-			}
 
-			group.Status = pkg.ComponentStatus(group.Summary.GetStatus())
```
yashmehrotra, line 368: Removing groups because it was creating an unnecessary subgroup. Also, we don't need summary and status since that is calculated at query time

PR #910, opened 2023-03-29
add logs selector fields in component

adityathebe:


api/v1/db_types.go
```diff
@@ -159,3 +159,47 @@ func (s Summary) GormValue(ctx context.Context, db *gorm.DB) clause.Expr {
...
+	var ba []byte
+	switch v := val.(type) {
+	case []byte:
+		ba = v
+	default:
+		return fmt.Errorf("value is not []byte: It's %T", val)
+	}
+
+	return json.Unmarshal(ba, rs)
+}
```
yashmehrotra, line 188: You can use duty's generic types like we do in incident commander
```
import github.com/flanksource/duty/types


func (p PersonProperties) Value() (driver.Value, error) {
	return types.GenericStructValue(p, true)
}

func (p *PersonProperties) Scan(val any) error {
	return types.GenericStructScan(&p, val)
}
```

Ref: https://github.com/flanksource/incident-commander/blob/3260cdebf9c498017df13715ff95f4adf174be85/api/team.go#L50-L56

moshloop:
@adityathebe can we move the logselectors type to duty/types ? and I think we also need to rename to json field into duty to just `logs`

adityathebe:
@moshloop renamed to `logs` in duty.

Do you mean we don't define logselector at all in here and directly use the `LogSelectors` struct from duty? I tried that and had some issues when running the `generate` script.

PR #931, opened 2023-04-06
feat: create checks from transformed check results

yashmehrotra:
Depends on https://github.com/flanksource/duty/pull/111

### DIFF
--- a/checks/alertmanager.go
+++ b/checks/alertmanager.go
@@ -63,7 +63,10 @@ func (c *AlertManagerChecker) Check(ctx *context.Context, extConfig external.Che
 
 	for alert := range alertMessage {
 		result := pkg.Success(check, ctx.Canary)
-		result.AddDetails(alertMessage[alert])
+		result.AddDetails(map[string]any{
+			"name":    alert,
+			"message": alertMessage[alert],
+		})
 		results = append(results, result)
 	}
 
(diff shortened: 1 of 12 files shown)


checks/alertmanager.go
```diff
@@ -63,7 +63,10 @@ func (c *AlertManagerChecker) Check(ctx *context.Context, extConfig external.Che
 
 	for alert := range alertMessage {
 		result := pkg.Success(check, ctx.Canary)
-		result.AddDetails(alertMessage[alert])
```
moshloop, line 66: I think we need to map in the labels and annotations - especially if we want to link alerts to nodes / namespaces / pods etc..

- We should also handle the starts at / ends at and status to map to check health

pkg/cache/postgres.go
```diff
@@ -34,10 +34,22 @@ func (c *postgresCache) Add(check pkg.Check, statii ...pkg.CheckStatus) {
...
 		c.AddCheckStatus(check, status)
 	}
 }
 
+func (c *postgresCache) AddCheckFromStatus(check pkg.Check, status pkg.CheckStatus) {
+	if status.Check == nil {
+		return
+	}
+	// Before syncing canary, mark all these checks as deleted_at
+	check.Labels["transformed-from"] = check.CanaryID.String()
```
moshloop, line 47: Is this always set even for non-transformed values ? It might be better to make this an actual column rather than a label

yashmehrotra: Have fixed the always set option, if you want I can make it a column.

moshloop: We probably also need to change it to not delete the check, but mark it as healthy / complete  - For alertmanager it might return the status as not firing and the time it stopped firing and this code would then only be used as a backup

PR #940, opened 2023-04-11
feat: topology metadata

adityathebe:
Resolves: https://github.com/flanksource/duty/issues/97

### DIFF
--- a/pkg/jobs/system/system_jobs.go
+++ b/pkg/jobs/system/system_jobs.go
@@ -10,6 +10,7 @@ import (
 	"github.com/flanksource/canary-checker/pkg"
 	"github.com/flanksource/canary-checker/pkg/db"
 	"github.com/flanksource/canary-checker/pkg/topology"
+	"github.com/flanksource/canary-checker/pkg/utils"
 	"github.com/flanksource/commons/logger"
 	"github.com/flanksource/duty/models"
 	"github.com/flanksource/kommons"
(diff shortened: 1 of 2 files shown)


pkg/api/topology.go
```diff
@@ -3,10 +3,62 @@ package api
...
+				exists = true
+				break
+			}
+		}
+
+		if !exists {
+			t.Tags = append(t.Tags, Tag{Key: k, Val: v})
+		}
+	}
+}
```
yashmehrotra, line 60: It would be faster if you just append tags and types and dedup them later in populate function

Right now, we will be iterating all the tags and all the types for all the components.

pkg/api/topology.go
```diff
@@ -3,10 +3,62 @@ package api
...
+	"github.com/flanksource/duty/models"
 	"github.com/labstack/echo/v4"
 )
 
+type Tag struct {
+	Key string `json:"key"`
+	Val string `json:"val"`
+}
+
+type TopologyRes struct {
```
yashmehrotra, line 17: Use `TopologyResponse`, do not use shortened names for structs unless it is required

pkg/api/topology.go
```diff
@@ -3,10 +3,62 @@ package api
...
+	componentIDs   []string          `json:"-"`
```
yashmehrotra, line 18: smaller case fields are not exported to json, the json tag can be removed

pkg/db/topology.go
```diff
@@ -15,6 +16,16 @@ import (
...
 )
 
+func GetTeamsOfComponents(ctx context.Context, componentIDs []string) ([]string, error) {
+	var teams []string
+	err := Gorm.WithContext(ctx).Table("team_components").Select("DISTINCT teams.name").Joins("LEFT JOIN teams ON teams.id = team_components.team_id").Where("component_id IN (?)", componentIDs).Find(&teams).Error
+	if err != nil {
+		return teams, err
+	}
+
+	return teams, nil
```
yashmehrotra, line 26:
```suggestion
	return teams, err
```

No need for the if check

yashmehrotra (changes requested):

adityathebe:
@yashmehrotra this is part of the API behavior, so I thought it's suited to be in canary-checker. Do you mean `topology.Query` should have this functionality?

pkg/db/topology.go
```diff
@@ -15,6 +16,12 @@ import (
 	"gorm.io/gorm/clause"
 )
 
+func GetTeamsOfComponents(ctx context.Context, componentIDs []string) ([]string, error) {
+	var teams []string
+	err := Gorm.WithContext(ctx).Table("team_components").Select("DISTINCT teams.name").Joins("LEFT JOIN teams ON teams.id = team_components.team_id").Where("component_id IN (?)", componentIDs).Find(&teams).Error
```
moshloop, line 21: I think is something that should be done as part of the main topology query

moshloop:
@adityathebe I think this she be part of duty ad part of the main topology query

adityathebe:
Addressed by https://github.com/flanksource/duty/pull/129

Issue #934, opened 2023-04-07
add a run now api


adityathebe:
similar to https://github.com/flanksource/config-db/issues/151

adityathebe:
@moshloop is there supposed to be 1 single endpoint that deals with both checks and system templates?

PR #941, opened 2023-04-12
feat: add run now handler for checks & system template

adityathebe:
Resolves: https://github.com/flanksource/canary-checker/issues/934

### DIFF
--- a/pkg/api/run_now.go
+++ b/pkg/api/run_now.go
@@ -9,6 +9,7 @@ import (
 	"github.com/flanksource/canary-checker/checks"
 	"github.com/flanksource/canary-checker/pkg"
 	"github.com/flanksource/canary-checker/pkg/db"
+	"github.com/flanksource/commons/logger"
 	"github.com/labstack/echo/v4"
 	"gorm.io/gorm"
 )
@@ -20,7 +21,25 @@ type RunNowRequest struct {
 
 // RunNowResponse represents the response body for a run now request
 type RunNowResponse struct {
-	Results []*pkg.CheckResult `json:"results"`
+	Total   int      `json:"total"`
+	Failed  int      `json:"failed"`
+	Success int      `json:"success"`
+	Errors  []string `json:"errors,omitempty"`
+}
+
+func (t *RunNowResponse) FromCheckResults(result []*pkg.CheckResult) {
+	t.Total = len(result)
+	for _, r := range result {
+		if r.Pass {
+			t.Success++
+			continue
+		}
+
+		t.Failed++
+		if r.Error != "" {
+			t.Errors = append(t.Errors, r.Error)
+		}
+	}
 }
 
 func RunNowHandler(c echo.Context) error {
@@ -32,7 +51,7 @@ func RunNowHandler(c echo.Context) error {
 	canaryModel, err := db.GetCanary(req.CanaryID)
 	if err != nil {
 		if errors.Is(err, gorm.ErrRecordNotFound) {
-			return errorResonse(c, fmt.Errorf("canary with id=%s was not found.", req.CanaryID), http.StatusNotFound)
+			return errorResonse(c, fmt.Errorf("canary with id=%s was not found", req.CanaryID), http.StatusNotFound)
 		}
 
 		return errorResonse(c, err, http.StatusBadRequest)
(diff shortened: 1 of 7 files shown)


pkg/api/run_now.go
```diff
@@ -0,0 +1,76 @@
...
+		return errorResonse(c, err, http.StatusInternalServerError)
+	}
+
+	kommonsClient, err := pkg.NewKommonsClient()
+	if err != nil {
+		logger.Warnf("failed to get kommons client, features that read kubernetes configs will fail: %v", err)
+	}
+
+	ctx := context.New(kommonsClient, *canary)
+	result := checks.RunChecks(ctx)
```
adityathebe, line 71: @moshloop Do we want to persist this as job history?

pkg/api/run_now.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/flanksource/canary-checker/checks"
+	"github.com/flanksource/canary-checker/pkg"
+	"github.com/flanksource/canary-checker/pkg/db"
+	"github.com/flanksource/commons/logger"
+	"github.com/labstack/echo/v4"
+	"gorm.io/gorm"
+)
+
+// RunNowRequest represents the request body for a run now request
+type RunNowRequest struct {
```
moshloop, line 18: Why can't we just use `/run/canary/{id}` and `/run/topology/{id}`

pkg/api/run_now.go
moshloop: No - Running now is not part of job history

pkg/db/canary.go
```diff
@@ -139,11 +139,9 @@ func DeleteChecks(id []string) error {
 func GetCanary(id string) (*pkg.Canary, error) {
 	var model *pkg.Canary
 	if err := Gorm.Where("id = ?", id).First(&model).Error; err != nil {
-		if errors.Is(err, gorm.ErrRecordNotFound) {
```
moshloop, line 142: Why do we remove this - callers of this method should not be aware of the implementation ? I also think we can switch to Find which never throws a record not  found

adityathebe: In general, FindByID db methods should either return the requested resource or return an error. A nil error means the resource was found. 

The previous implementation can potentially return 2 nil values which, in my opinion, is a bad practice as that can cause accidental panics. This requires 2 nil checks for all the callers.

I agree that callers needing to check against the gorm error is not a good practice. Ideally, we'd need a domain level "not found" error and the db methods would map `gorm.ErrRecordNotFound` error to our native NotFound error. But, we currently don't have such domain errors.

What do you think @moshloop

pkg/api/run_now.go
adityathebe: Sure we can. The requirement wasn't clear whether we needed a single handler or 2 separate.

pkg/topology/run.go
```diff
@@ -406,3 +407,54 @@ func Run(opts TopologyRunOptions, s v1.SystemTemplate) []*pkg.Component {
...
+		if err := db.DeleteComponentsWithIDs(deleteCompIDs, time.Now()); err != nil {
+			return fmt.Errorf("failed to delete components: %w", err)
+		}
+	}
+
+	return nil
+}
+
+// difference returns the elements in `a` that aren't in `b`.
+func difference(a, b []uuid.UUID) []string {
```
yashmehrotra, line 448: We already have a utils.SetDifference for this

Can you convert it into a generic function which takes `T comparable` ?

pkg/db/canary.go
yashmehrotra: I usually prefer:
Find -> can or cannot exist, not found should not return err
Get -> has to exist, not found raises an error

For fetch,some argue it should work like find and others argue it should work like get, I'd prefer we only use Find and Get  not use Fetch

Ref: https://dev.to/maikomiyazaki/beginner-s-cheat-sheet-commonly-used-verbs-for-naming-functions-methods-and-variables-509i

adityathebe: @yashmehrotra That's a new one to me. Personally I would prefer to not have to deal with nuances like that. `Find, Get & Fetch` are I think used pretty much interchangeably. And, a method that conveys that the resource was not found either through errors or nil value should be sufficient to deal with any use cases.

@moshloop I guess both approaches inform the user that the resource was not found in different ways but I still think a `not found` error is more suitable simply to prevent accidentally accessing nil values after err check.

yashmehrotra: > Find, Get & Fetch are I think used pretty much interchangeably
Have to disagree here, Find and Get are fundamentally different on a semantic grammar level, you can "Find" something but you may or may not have what you were looking for. When you "Get" something, the implication is that it already exists

adityathebe: I'm not denying that there's a semantic difference between them. I'm just saying those are not standard verbs and are used synonymously in practice - even in our codebase.

Anyway, don't want to take too much time on this one. I have made changes.

PR #959, opened 2023-04-27
feat: connection support

adityathebe:
Resolves: https://github.com/flanksource/canary-checker/issues/952

Also resolves: https://github.com/flanksource/duty/issues/77


checks/checker.go
```diff
@@ -24,34 +24,33 @@ type Checker interface {
...
+	&MssqlChecker{},
+	&MysqlChecker{},
+	&PostgresChecker{},
+	&PrometheusChecker{},
+	&RedisChecker{},
+	&ResticChecker{},
+	&S3Checker{},
 	NewNamespaceChecker(),
+	NewPodChecker(),
 	NewTCPChecker(),
```
adityathebe, line 55: No new change here. Just sorted it alphabetically.

api/v1/checks.go
```diff
@@ -1138,38 +1313,38 @@ func (c EC2Check) GetType() string {
...
+	MysqlCheck{},
+	NamespaceCheck{},
+	PodCheck{},
+	PostgresCheck{},
+	PrometheusCheck{},
+	RedisCheck{},
+	ResticCheck{},
+	S3Check{},
+	TCPCheck{},
 }
```
adityathebe, line 1350: No new change here. Just sorted it alphabetically.

api/v1/checks.go
```diff
@@ -1138,38 +1345,38 @@ func (c EC2Check) GetType() string {
...
 
 var AllChecks = []external.Check{
```
adityathebe, line 1347: No new changes here. Just sorted it alphabetically.

checks/checker.go
```diff
@@ -24,34 +24,33 @@ type Checker interface {
 }
 
 var All = []Checker{
```
adityathebe, line 26: No new changes here. Just sorted it alphabetically.

api/v1/checks.go
```diff
@@ -220,6 +229,30 @@ func (c JmeterCheck) GetType() string {
 	return "jmeter"
 }
 
+// PopulateConnection will attempt to populate the host and port from the connection name.
+func (c *JmeterCheck) PopulateConnection(ctx context.Context, db *gorm.DB, k8sClient kubernetes.Interface, namespace string) error {
+	if c.ConnectionName == "" {
+		return nil
+	}
+
+	connection, err := duty.HydratedConnectionByURL(ctx, db, k8sClient, namespace, c.ConnectionName)
```
moshloop, line 238: The method name should be `HydrateConnectionByURL`

also not sure if this method should be on the API itself  - @yashmehrotra wdyt ?

api/v1/checks.go
```diff
@@ -220,6 +229,30 @@ func (c JmeterCheck) GetType() string {
 	return "jmeter"
 }
 
+// PopulateConnection will attempt to populate the host and port from the connection name.
+func (c *JmeterCheck) PopulateConnection(ctx context.Context, db *gorm.DB, k8sClient kubernetes.Interface, namespace string) error {
```
moshloop, line 233:
```suggestion
func (c *JmeterCheck) HydrateConnection(ctx context.Context, db *gorm.DB, k8sClient kubernetes.Interface, namespace string) error {
```

api/v1/checks.go
```diff
@@ -527,6 +581,7 @@ func (c NamespaceCheck) GetType() string {
 
 type DNSCheck struct {
 	Description     `yaml:",inline" json:",inline"`
+	ConnectionName  string   `yaml:"connection,omitempty" json:"connection,omitempty"`
```
moshloop, line 584: I don't think we need this - DNS servers are not authenticated

api/v1/checks.go
```diff
@@ -628,13 +686,86 @@ func (c SMBConnection) GetPort() int {
...
+	connection, err := duty.HydratedConnectionByURL(ctx, db, k8sClient, namespace, c.ConnectionName)
+	if err != nil {
+		return false, err
+	}
+
+	if connection == nil {
+		return false, nil
+	}
+
+	c.Domain = connection.URL
```
moshloop, line 703: Domain is something that comes from the username e.g. johndoe@acme.corp and the domain is acme.corp

api/v1/checks.go
```diff
@@ -755,9 +892,58 @@ type AWSConnection struct {
...
 
+// populateFromConnection attempts to find the connection by name
+// and populate the endpoint, accessKey and secretKey.
+func (t *AWSConnection) populateFromConnection(ctx context.Context, db *gorm.DB, k8sClient kubernetes.Interface, namespace string) error {
+	connection, err := duty.HydratedConnectionByURL(ctx, db, k8sClient, namespace, t.ConnectionName)
+	if err != nil {
+		return err
+	}
+
+	if connection == nil {
```
moshloop, line 903: Isn't this an expected condition = if connectionName != "" and connection == nil && err == nil. then is connection not found, but if it is connectionName == "" then its a noop

api/v1/checks.go
```diff
@@ -755,9 +892,58 @@ type AWSConnection struct {
...
+	t.SecretKey.Value = connection.Password
+	t.Endpoint = connection.URL
+
+	return nil
+}
+
+// Populate populates an AWSConnection with credentials and other information.
+// If a connection name is specified, it'll be used to populate the endpoint, accessKey and secretKey.
+func (t *AWSConnection) Populate(ctx context.Context, db *gorm.DB, kommonsClient *kommons.Client, namespace string) error {
+	if t.ConnectionName == "" {
```
moshloop, line 917: Isn't this meant to be a != ?

api/v1/checks.go
moshloop: I also don't think we need to pass through all the args, we can just pass through context, and then add a context.HydrateConnectionByURL() method that passes through to duty after initializing db / k8s 

If we make this also an interface, we can just call the method on every check if it matches the interface.

api/v1/checks.go
adityathebe: Ya, that's why I only used `duty.FindConnectionByURL` instead of `duty.HydratedConnectionByURL` because there's no auth. Thought it would still be handy to just get the host from a connection name.


But, i'll remove it.

api/v1/checks.go
adityathebe: There was a typo in `.Populate()` . I think that should address this.

api/v1/checks.go
adityathebe: @moshloop I think we should probably do it while we're saving the connection.

api/v1/checks.go
adityathebe: > The method name should be HydrateConnectionByURL

This, to me, implies that you already have a connection and you want to hydrate it.

```go
func HydrateConnectionByURL(ctx context.Context, connection Connection) {
    ...
}
```

checks/restic.go
```diff
@@ -45,18 +45,41 @@ func (c *ResticChecker) Check(ctx *context.Context, extConfig external.Check) pk
...
+
+		envVars = map[string]string{
+			resticPasswordEnvKey: connection.Password,
+			// resticAwsSecretAccessKey:   secretKey,
+			// resticAwsAccessKeyIDEnvKey: connection,
+		}
+
+		// TODO: Not sure why wee need AWS credentials here.
+		// If it's required, will need to think how a single connection can provide
+		// both AWS credentials and a restic password.
```
adityathebe, line 65: @moshloop

checks/aws_config.go
```diff
@@ -34,13 +34,20 @@ func (c *AwsConfigChecker) Check(ctx *context.Context, extConfig external.Check)
 	result := pkg.Success(check, ctx.Canary)
 	var results pkg.Results
 	results = append(results, result)
+
 	if check.AWSConnection == nil {
 		check.AWSConnection = &v1.AWSConnection{}
+	} else {
+		if err := check.AWSConnection.Populate(ctx, ctx.Kommons, ctx.Namespace); err != nil {
```
moshloop, line 41: Is this not a nil pointer ?

checks/azure_devops.go
```diff
@@ -40,30 +39,22 @@ func (t *AzureDevopsChecker) check(ctx *context.Context, check v1.AzureDevopsChe
 	var personalAccessToken string
 	if check.PersonalAccessToken.ValueStatic != "" {
 		personalAccessToken = check.PersonalAccessToken.ValueStatic
+	} else if connection, err := ctx.HydrateConnectionByURL(check.ConnectionName); err != nil {
+		return results.Failf("failed to hydrate connection: %v", err)
+	} else if connection != nil {
+		personalAccessToken = connection.Password
```
moshloop, line 45: I don't think this will ever execute

checks/folder_sftp.go
```diff
@@ -14,15 +14,24 @@ func CheckSFTP(ctx *context.Context, check v1.FolderCheck) pkg.Results {
 	result := pkg.Success(check, ctx.Canary)
 	var results pkg.Results
 	results = append(results, result)
-	auth, err := GetAuthValues(check.SFTPConnection.Auth, ctx.Kommons, ctx.Canary.Namespace)
+
+	foundConn, err := check.SFTPConnection.HydrateConnection(ctx)
 	if err != nil {
-		return results.ErrorMessage(err)
+		return results.Failf("failed to populate SFTP connection: %v", err)
```
moshloop, line 20: Why is this logic different from the others ?

checks/github.go
```diff
@@ -36,16 +37,31 @@ func (c *GitHubChecker) Check(ctx *context.Context, extConfig external.Check) pk
...
 	results = append(results, result)
-	_, githubToken, err := ctx.Kommons.GetEnvValue(*check.GithubToken, ctx.Canary.GetNamespace())
-	if err != nil {
-		return results.Failf("error fetching github token: %v", err)
+
+	var githubToken string
+	if connection, err := ctx.HydrateConnectionByURL(check.ConnectionName); err != nil {
+		return results.Failf("failed to find connection for github token %q: %v", check.ConnectionName, err)
+	} else if connection != nil {
+		githubToken = connection.Password
```
moshloop, line 45: this will never run

checks/http.go
```diff
@@ -101,18 +101,23 @@ func truncate(text string, max int) string {
...
 	check := extConfig.(v1.HTTPCheck)
 	var results pkg.Results
+	var err error
 	result := pkg.Success(check, ctx.Canary)
 	results = append(results, result)
+
+	if connection, err := ctx.HydrateConnectionByURL(check.ConnectionName); err != nil {
+		return results.Failf("failed to find HTTP connection %q: %v", check.ConnectionName, err)
+	} else if connection != nil {
+		check.Endpoint = connection.URL
```
moshloop, line 114: this will never run

checks/ldap.go
```diff
@@ -35,20 +35,26 @@ func (c *LdapChecker) Check(ctx *context.Context, extConfig external.Check) pkg.
...
 	var results pkg.Results
+	var err error
 	results = append(results, result)
-	ld, err := ldap.DialURL(check.Host, ldap.DialWithTLSConfig(&tls.Config{
-		InsecureSkipVerify: check.SkipTLSVerify,
-	}))
-	if err != nil {
-		return results.Failf("Failed to connect %v", err)
+
+	if ok, err := check.HydrateConnection(ctx); err != nil {
```
moshloop, line 41: Can we make this follow the ctx.HydrateConnection pattern

checks/restic.go
moshloop: Restic stores its backups in AWS S3

api/v1/checks.go
```diff
@@ -105,6 +115,7 @@ func (t TCPCheck) GetType() string {
 type ICMPCheck struct {
 	Description         `yaml:",inline" json:",inline"`
 	Endpoint            string `yaml:"endpoint" json:"endpoint,omitempty"`
+	ConnectionName      string `yaml:"connection,omitempty" json:"connection,omitempty"`
```
moshloop, line 118: This is not needed

api/v1/checks.go
```diff
@@ -125,17 +136,13 @@ type Bucket struct {
...
-	Description `yaml:",inline" json:",inline"`
-	Bucket      Bucket `yaml:"bucket" json:"bucket,omitempty"`
-	AccessKey   string `yaml:"accessKey" json:"accessKey,omitempty"`
-	SecretKey   string `yaml:"secretKey" json:"secretKey,omitempty"`
-	ObjectPath  string `yaml:"objectPath" json:"objectPath,omitempty"`
-	// Skip TLS verify when connecting to s3
-	SkipTLSVerify bool `yaml:"skipTLSVerify,omitempty" json:"skipTLSVerify,omitempty"`
+	Description   `yaml:",inline" json:",inline"`
+	AWSConnection `yaml:",inline" json:",inline"`
+	BucketName    string `yaml:"bucketName" json:"bucketName,omitempty"`
```
moshloop, line 141: Why the rename of the field ?

api/v1/checks.go
moshloop: ok, then pull domain from the properties

api/v1/checks.go
```diff
@@ -649,9 +793,11 @@ func (c SFTPConnection) GetPort() int {
 type Prometheus struct {
 	PrometheusCheck `yaml:",inline" json:",inline"`
 }
+
 type PrometheusCheck struct {
-	Description `yaml:",inline" json:",inline"`
-	Templatable `yaml:",inline" json:",inline"`
+	Description    `yaml:",inline" json:",inline"`
+	Templatable    `yaml:",inline" json:",inline"`
+	ConnectionName string `yaml:"connection" json:"connection"`
```
moshloop, line 800: Missing omitempty

api/v1/checks.go
```diff
@@ -682,11 +842,12 @@ type Git struct {
 }
 
 type GitHubCheck struct {
-	Description `yaml:",inline" json:",inline"`
-	Templatable `yaml:",inline" json:",inline"`
+	Description    `yaml:",inline" json:",inline"`
+	Templatable    `yaml:",inline" json:",inline"`
+	ConnectionName string `yaml:"connection,inline" json:"connection,inline"`
```
moshloop, line 847: Missing omitEmpty and remove inline

api/v1/checks.go
```diff
@@ -756,9 +920,41 @@ type AWSConnection struct {
...
+	}
+
+	return nil
+}
+
 type GCPConnection struct {
-	Endpoint    string          `yaml:"endpoint" json:"endpoint,omitempty"`
-	Credentials *kommons.EnvVar `yaml:"credentials" json:"credentials,omitempty"`
+	// ConnectionName of the connection. It'll be used to populate the endpoint and credentials.
+	// It's named "connection" instead of just "name" so that it fits with the convention.
```
moshloop, line 954: Why do we need this comment ?

checks/aws_config.go
adityathebe: I don't think so.

api/v1/checks.go
adityathebe: Removed connection from ICMP

api/v1/checks.go
adityathebe: Because the structure itself has changed. Before "bucket" was a struct

```yaml
  s3:
    - bucket:
        name: "test-bucket"
        region: "us-east-1"
        endpoint: "https://test-bucket.s3.us-east-1.amazonaws.com"
      secretKey: "****************"
      accessKey: "~~~~~~~~~~~~~~~~"
      objectPath: "path/to/object"
      skipTLSVerify: true
```

but I changed it so AWSConnection can be used like we do for all other AWS related checks.

I can rename to `bucket` but a change to YAML files are needed regardless.

checks/azure_devops.go
adityathebe: It'll execute if the connection name is empty.

checks/github.go
adityathebe: Why not if it find the connection?

checks/folder_sftp.go
adityathebe: If you mean why it's using `Failf` instead of `ErrorMessage()` then it's for no particular reason. I prefer `Failf` as it adds more context.

checks/ldap.go
adityathebe: `check.HydrateConnection` is internally doing that [here](https://github.com/flanksource/canary-checker/pull/959/files/217daf66778c68ca87905f91857d095a216ecba4#diff-b2b2311aec83c084649791c6470895a8e34180b9ee76afbae9d9efdfb598303cR549)

checks/restic.go
adityathebe: okay. Password is already taken so I'll use the properties to get the AWS keys.

adityathebe: Create new field for aws connection

Issue #968, opened 2023-05-03
Handle bad canary spec


yashmehrotra:
All the canaries in the db are parsed together as a json list, if any spec is bad, the entire parsing gets broken and an error is returned

![Screenshot from 2023-05-03 08-56-32](https://user-images.githubusercontent.com/5863972/235899581-9744b61f-59af-4278-a88e-f466b368e5a8.png)

Issue #969, opened 2023-05-03
Health checks added via the UI do not run


moshloop:
A failed job history should be added for any spec marshalling / setup errors

PR #971, opened 2023-05-04
fix: handle invalid canary parsing

yashmehrotra:
Fixes: #968 and #969


pkg/db/canary.go
```diff
@@ -48,7 +48,8 @@ func GetAllCanaries() ([]v1.Canary, error) {
 	for _, _canary := range _canaries {
 		c, err := _canary.ToV1()
 		if err != nil {
-			return nil, err
+			logger.Errorf("Error parsing canary[%s]: %v", _canary.ID, err)
```
moshloop, line 51: @yashmehrotra  Can we add a job history event as well ?

yashmehrotra: Have added

PR #1000, opened 2023-05-18
feat: generate minimal and elevated base images

BrendanGalloway:


moshloop:
@BrendanGalloway can we update the helm chart as well to provide a bool flip between the images

build/full/Dockerfile
```diff
@@ -3,6 +3,7 @@ WORKDIR /app
 
 ARG NAME
 ARG VERSION
+ENV IMAGE_TYPE=elevated
```
moshloop, line 6:
```suggestion
ENV IMAGE_TYPE=full
```

PR #1003, opened 2023-05-19
chore: change git test params

yashmehrotra:


fixtures/git/git_check_pass.yaml
```diff
@@ -8,7 +8,9 @@ spec:
     - query: "SELECT * FROM github_repo_checks('flanksource/canary-checker') where branch='master' and conclusion='FAILURE'"
       name: github-check
       test:
-        expr: len(results) == 0
+        expr: len(results) != 0
```
moshloop, line 11: Maybe we should remove the conclusion and just have len(results) > 0

Issue #1014, opened 2023-05-22
Wrong namespace being populated for pods


yashmehrotra:
The component namespace is set as the namespace of topology instead of the pod namespace. We do set the namespace in pod's component object but that seems to get overridden somewhere.

PR #1015, opened 2023-05-22
fix: component namespace should take precedence over topology namespace

yashmehrotra:
Fixes: https://github.com/flanksource/canary-checker/issues/1014


yashmehrotra:
@moshloop FYI

PR #1033, opened 2023-05-29
chore: set serviceMonitor default to false

BrendanGalloway:


moshloop:
@BrendanGalloway  can we update the READMe  as well ?

Issue #1028, opened 2023-05-25
Pod check is not failing


moshloop:
The Pod check currently has this message: `request timeout exceeded 0ms > 7000`

The canary should in fact fail,  as the domain is not correct. 

https://mission-control.azure.lab.flanksource.com/health?hidePassing=false&sortBy=name&sortDesc=false&labels=%257B%257D&query=pod&layout=table&checkId=0188432c-ef26-2051-594d-eb7684af3da4&timeRange=1h

yashmehrotra:
If ingress and host are empty strings, use service name to check

PR #1036, opened 2023-05-30
fix: handle empty ingress config for pod check 

yashmehrotra:
Fixes: https://github.com/flanksource/canary-checker/issues/1028


moshloop:
@yashmehrotra can we add a pod_fail fixture for a container does fails to serve traffic ?

Issue #1013, opened 2023-05-21
Alertmanager alerts should be unique


moshloop:
Not 1 per type, But 1 per firing instance

yashmehrotra:
Order to append:
namespace + pod (xyz/<pod_name>)
Order deployment/daemonset/pod

PR #1037, opened 2023-05-30
feat: add name suffix for alertmanager check

yashmehrotra:
Fixes: https://github.com/flanksource/canary-checker/issues/1013

Names will now look like this:

```
+-------------------------------------------------------------+
| name                                                        |
|-------------------------------------------------------------|
| alertmanager-check                                          |
| KubePodCrashLooping-dev/backend-admin-app-6c4f67756c-sld95  |
| KubeStateMetricsListErrors                                  |
| KubePodNotReady-default/k8s-check-not-ready                 |
| KubeJobFailed-demo/create-postgres-db                       |
| Watchdog                                                    |
| KubeletTooManyPods-ip-10-0-6-69.eu-west-2.compute.internal  |
| KubePodCrashLooping-prod/public-website-777dbd4c57-96wdl    |
| KubePodCrashLooping-prod/backend-admin-app-7b58d4f6b7-h4629 |
| KubeDeploymentReplicasMismatch-dev/public-website           |
| KubePodCrashLooping-dev/public-website-694767fbff-5bbfn     |
| TargetDown                                                  |
| KubeDeploymentReplicasMismatch-platform-system/karina       |
| KubeJobNotCompleted                                         |
| KubeContainerWaiting-default/k8s-check-not-ready/hello      |
+-------------------------------------------------------------+

```

### DIFF
--- a/checks/alertmanager.go
+++ b/checks/alertmanager.go
@@ -67,7 +67,7 @@ func (c *AlertManagerChecker) Check(ctx *context.Context, extConfig external.Che
 	var alertMessages []map[string]any
 	for _, alert := range alerts.Payload {
 		alertMap := map[string]any{
-			"name":        generateNameSuffix(alert.Labels["alertname"], alert.Labels),
+			"name":        generateFullName(alert.Labels["alertname"], alert.Labels),
 			"message":     extractMessage(alert.Annotations),
 			"labels":      alert.Labels,
 			"annotations": alert.Annotations,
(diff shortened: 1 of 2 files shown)


checks/alertmanager.go
```diff
@@ -88,3 +89,25 @@ func extractMessage(annotations map[string]string) string {
...
+		"KubeletTooManyPods":             {"node"},
+		"KubeContainerWaiting":           {"namespace", "pod", "container"},
+	}
+
+	if val, exists := nameSuffixFromLabels[name]; exists {
+		var suffix []string
+		for _, v := range val {
+			suffix = append(suffix, labels[v])
+		}
+		return name + "-" + strings.Join(suffix, "/")
```
moshloop, line 110:
```suggestion
		return name + "/" + strings.Join(suffix, "/")
```

checks/alertmanager.go
```diff
@@ -88,3 +89,25 @@ func extractMessage(annotations map[string]string) string {
...
 	return ""
 }
+
+func generateNameSuffix(name string, labels map[string]string) string {
+	// Based on the type of alert, we will add alert metadata to the check name
+	// The key here is the alert type and the name will be "/".join(value)
+	nameSuffixFromLabels := map[string][]string{
+		"KubePodCrashLooping":            {"namespace", "pod"},
+		"KubePodNotReady":                {"namespace", "pod"},
+		"KubeDeploymentReplicasMismatch": {"namespace", "deployment"},
```
moshloop, line 99: Can we make the heuristic and not depend on the alert name? Also need to account for Daemonset and Statefulset

checks/alertmanager.go
yashmehrotra: @moshloop This would lead to the name being `KubePodCrashLooping/prod/backend-admin-app-7b58d4f6b7-h4629` instead of `KubePodCrashLooping-prod/backend-admin-app-7b58d4f6b7-h4629`

Do we want that ? There is no distinction b/w alert type and id then

checks/alertmanager.go
```diff
@@ -88,3 +89,25 @@ func extractMessage(annotations map[string]string) string {
...
+		"KubeJobFailed":                  {"namespace", "job_name"},
+		"KubeletTooManyPods":             {"node"},
+		"KubeContainerWaiting":           {"namespace", "pod", "container"},
+	}
+
```
yashmehrotra, line 104: namespace/ 1st of deployment

checks/alertmanager.go
moshloop: I think it reads better with a `/` rather than 2 different seperators

checks/alertmanager.go
moshloop: I think this will create a very long name,

it should be : 

If namespace exists, add it
Add first label from priority list

PR #1050, opened 2023-05-31
fix: set default permissions for github workflows as read only

dabasvibhor:


.github/workflows/release.yml
```diff
@@ -1,4 +1,6 @@
 name: Create Release
+# Declare default permissions as read only
+permissions: read-all
```
moshloop, line 3: @dabasvibhor I think the release workflow will need elevated privileges

yashmehrotra (approved):

Issue #610, opened 2022-09-14
Kubernetes topology v2


moshloop:
- [ ] Add cpu/mem/storage limits 
    -  For nodes based on allocatable, 
     - For pods based on limits 
- Add cluster level cpu/memory metric

PR #1076, opened 2023-06-15
fix: reference error in property lookup and update k8s topology

yashmehrotra:
Fixes: https://github.com/flanksource/canary-checker/issues/610

### DIFF
--- a/fixtures/topology/k8s-system.yaml
+++ b/fixtures/topology/k8s-system.yaml
@@ -44,15 +44,32 @@ spec:
                 name: k8s
                 display:
                   javascript: JSON.stringify(k8s.getNodeTopology(results)) 
-          properties:
-            - name: node-metrics
-              lookup:
-                kubernetes:
-                  - kind: NodeMetrics
-                    ready: false
-                    name: nodemetrics
+          forEach:
+            properties:
+              - name: CPU %
+                lookup:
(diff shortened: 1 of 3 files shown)


templating/js/k8s.js
```diff
@@ -315,10 +330,19 @@ k8s = {
             headline: true,
             max: fromSI(node.status.allocatable.memory)
           },
+          {
+            name: "memory limit",
+            unit: "gigabytes",
+            text: (fromSI(node.status.allocatable.memory) / (1024 * 1024 * 1024)).toString(),
```
moshloop, line 336: this should be max in the memory property

fixtures/topology/k8s-system.yaml
```diff
@@ -44,15 +44,32 @@ spec:
...
-                kubernetes:
-                  - kind: NodeMetrics
-                    ready: false
-                    name: nodemetrics
+          forEach:
+            properties:
+              - name: cpu
+                lookup:
+                  prometheus:
+                  - query: '1000 * max(rate(container_cpu_usage_seconds_total{container!="",node="${.component.name}"}[5m]))'
```
moshloop, line 52: This shouldn't be a forEach which is N+1, if you remove the node filter and return a map of lists then we only need to execute a single prometheus query

PR #1086, opened 2023-06-20
feat: add a new endpoint to delete canary

adityathebe:


pkg/db/canary.go
```diff
@@ -113,33 +113,60 @@ func UpdateChecksStatus(ids []string, status models.CheckHealthStatus) error {
...
+	return nil
+}
+
+// DeleteCanaryByID deletes a canary and the associated checks and relationships.
+func DeleteCanaryByID(canaryID string) error {
+	if err := unregisterChecksGauge(canaryID); err != nil {
+		logger.Errorf("failed to unregister checks gauges: %v", err)
+	}
+
+	return deleteCanaryByID(canaryID)
```
yashmehrotra, line 151: I think we should delete canary first then unregister gauge

If gauge gets unregistered, but canary deletion fails, do they get re-synced up ?

pkg/db/canary.go
```diff
@@ -113,33 +113,60 @@ func UpdateChecksStatus(ids []string, status models.CheckHealthStatus) error {
...
-	}
-	if err := DeleteChecksForCanary(persistedID, deleteTime); err != nil {
-		return err
-	}
-	if err := DeleteCheckComponentRelationshipsForCanary(persistedID, deleteTime); err != nil {
-		return err
-	}
-	return nil
+
+	return deleteCanaryByID(persistedID)
```
yashmehrotra, line 169: (nit): just use `id`

pkg/api/api.go
```diff
@@ -34,6 +36,25 @@ func About(c echo.Context) error {
...
+func DeleteCanary(c echo.Context) error {
+	id := c.Param("id")
+
+	canary, err := db.FindCanaryByID(id)
+	if err != nil {
+		return errorResonse(c, err, http.StatusInternalServerError)
+	} else if canary == nil {
+		return errorMsgResponse(c, http.StatusNotFound, "canary with id=%s was not found", id)
+	}
+
```
yashmehrotra, line 48: Is this required ? We are just checking if the canary with ID exists or not, but the same can happen in the db.DeleteCanaryByID where it should return a `record not found` if we pass the id directly

pkg/db/canary.go
adityathebe: > If gauge gets unregistered, but canary deletion fails, do they get re-synced up ?

If the canary deletion fails and the checks keep on running then the gauge is re-registered.

pkg/api/api.go
adityathebe: I did this so the HTTP error is well formed _(with the correct status and message)_.

Or else I'd have to either check the gorm `ErrRecordNotFound` error in pkg/api or just transfer the error message from `db.DeleteCanaryByID` with incorrect HTTP status.

moshloop:
@adityathebe do we still need this if we are doing cascading deletes?

We can probably use a trigger to have the deleted_at columns propagate to check_status and check_relationships ?

adityathebe:
> @adityathebe do we still need this if we are doing cascading deletes?
> 
> We can probably use a trigger to have the deleted_at columns propagate to check_status and check_relationships ?

@moshloop I think we still need the endpoint to
- remove the metrics, and
- stop the health check jobs?

adityathebe:
We decided to not create a new endpoint to delete canaries. Instead, we will be 
- adding triggers to handle soft deletes
- cascading deletes to handle hard deletes

The metric gauges will be cleared up eventually by the cleaning job and if we want it to be real-time, then we can publish a new event and act upon that.

The health check jobs are effectively stopped due to the deleted at check.

PR #1094, opened 2023-06-23
fix: mark transformed checks as deleted

yashmehrotra:
Fixes: https://github.com/flanksource/canary-checker/issues/1093

### DIFF
--- a/pkg/db/canary.go
+++ b/pkg/db/canary.go
@@ -109,7 +109,10 @@ func UpdateChecksStatus(ids []string, status models.CheckHealthStatus) error {
 	}
 	return Gorm.Table("checks").
 		Where("id in (?)", ids).
-		Update("status", status).
+		Updates(map[string]any{
+			"status":     status,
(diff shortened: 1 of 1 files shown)


pkg/db/canary.go
```diff
@@ -109,7 +109,10 @@ func UpdateChecksStatus(ids []string, status models.CheckHealthStatus) error {
 	}
 	return Gorm.Table("checks").
 		Where("id in (?)", ids).
-		Update("status", status).
+		Updates(map[string]any{
+			"status":     status,
+			"deleted_at": time.Now(),
```
moshloop, line 114: shouldnt this be server now ?

yashmehrotra: time.Now() also has tz context, so it should be the same

moshloop: Not if there is a clock skew between db and server

yashmehrotra: 👍🏽

PR #1099, opened 2023-06-26
fix: unnecessary upsert and dirty data for checks

yashmehrotra:


pkg/cache/postgres.go
```diff
@@ -51,6 +51,10 @@ func (c *postgresCache) AddCheckFromStatus(check pkg.Check, status pkg.CheckStat
 		return uuid.Nil, nil
 	}
 
+	if !check.Transformed {
```
yashmehrotra, line 54: This flow is only for transformed checks but some untransformed checks were being processed as well, causing extra upserts

pkg/db/canary.go
```diff
@@ -69,19 +77,31 @@ func PersistCheck(check pkg.Check, canaryID uuid.UUID) (uuid.UUID, error) {
 	}
 	check.Name = name
 	check.Description = description
+
+	// TODO: Find root cause why pod check has these labels in check model
+	delete(check.Labels, "canary-checker.flanksource.com/podCheck")
+	delete(check.Labels, "canary-checker.flanksource.com/check")
```
yashmehrotra, line 83: These labels are added by the pod check run, but they keep on changing leading to excessive upserts. Adding a TODO for the root cause currently

pkg/db/canary.go
```diff
@@ -261,6 +281,7 @@ func PersistCanary(canary v1.Canary, source string) (*pkg.Canary, map[string]str
 		if checkID := canary.GetCheckID(check.Name); checkID != "" {
 			check.ID, _ = uuid.Parse(checkID)
 		}
+		check.Spec, _ = json.Marshal(config)
```
yashmehrotra, line 284: Moving spec creation here guarantees desired behavior

PR #1103, opened 2023-06-28
feat: do not run changelog trigger migration

yashmehrotra:
Fixes: https://github.com/flanksource/canary-checker/issues/1102

### DIFF
--- a/pkg/db/init.go
+++ b/pkg/db/init.go
@@ -11,9 +11,8 @@ import (
 	embeddedpostgres "github.com/fergusstrange/embedded-postgres"
 	"github.com/flanksource/commons/logger"
 	"github.com/flanksource/duty"
-	"github.com/jackc/pgx/v5"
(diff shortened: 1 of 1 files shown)


pkg/db/init.go
```diff
@@ -120,45 +119,14 @@ func Init() error {
...
-		if err := duty.Migrate(ConnectionString, nil); err != nil {
+		if err := duty.Migrate(ConnectionString, &migrate.MigrateOptions{IgnoreFiles: []string{"012_changelog.sql"}}); err != nil {
 			return err
 		}
 	}
 
 	return nil
 }
 
-func Cleanup() {
```
moshloop, line 131: Where does this happen now ?

yashmehrotra: This func was never being called, creating a new PR for cleaning up old checks

Issue #1004, opened 2023-05-19
Prevent multiple simultaneous execution of the same check


adityathebe:
**Example**: Assuming there's an HTTP check running every 30 sec and the the HTTP endpoint returns after 5 minutes on average. There'll be 9 simultaneous job running for the same check before the first one completes.

The current `if lastRunDelta < minimumTimeBetweenCanaryRuns` check doesn't prevent this.

PR #1108, opened 2023-06-29
feat: prevent concurrent execution of the same canary job

adityathebe:
Resolves: https://github.com/flanksource/canary-checker/issues/1004

### DIFF
--- a/pkg/jobs/canary/suite_test.go
+++ b/pkg/jobs/canary/suite_test.go
@@ -47,8 +47,7 @@ func DelayedResponseHandler(c echo.Context) error {
 	}
 
 	time.Sleep(time.Duration(delay) * time.Second)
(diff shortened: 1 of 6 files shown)


pkg/jobs/canary/suite_test.go
```diff
@@ -0,0 +1,142 @@
...
+		ginkgo.Fail(err.Error())
+	}
+
+	logger.Infof("Stopping postgres")
+	if err := postgresServer.Stop(); err != nil {
+		ginkgo.Fail(err.Error())
+	}
+})
+
+func setupDB(connectionString string) (*gorm.DB, *pgxpool.Pool) {
```
adityathebe, line 94: We could move this to commons since it'll be used in pretty much all repos

PR #1111, opened 2023-06-29
chore: inline auth structs

moshloop:
Made all authentication inline to provide consistency


api/v1/checks.go
```diff
@@ -1011,7 +971,17 @@ func (t *AWSConnection) Populate(ctx checkContext, k8s kubernetes.Interface, nam
 
 		t.AccessKey.ValueStatic = connection.Username
 		t.SecretKey.ValueStatic = connection.Password
-		t.Endpoint = connection.URL
+		if t.Endpoint == "" {
+			t.Endpoint = connection.URL
+		}
```
adityathebe, line 976: Why do we partially use the connection only in here?

api/v1/checks.go
```diff
@@ -532,12 +509,13 @@ func (c *ElasticsearchCheck) HydrateConnection(ctx checkContext) error {
...
 type DynatraceCheck struct {
-	Description `yaml:",inline" json:",inline"`
-	Templatable `yaml:",inline" json:",inline"`
-	Host        string       `yaml:"host" json:"host,omitempty" template:"true"`
-	Scheme      string       `yaml:"scheme" json:"scheme,omitempty"`
-	APIKey      types.EnvVar `yaml:"apiKey" json:"apiKey,omitempty"`
-	Namespace   string       `yaml:"namespace" json:"namespace,omitempty" template:"true"`
+	Description    `yaml:",inline" json:",inline"`
+	Templatable    `yaml:",inline" json:",inline"`
+	ConnectionName string       `yaml:"connection,omitempty" json:"connection,omitempty"`
```
adityathebe, line 514: Connection name is added but it's not being used

config/schemas/topology.schema.json
```diff
@@ -569,14 +612,15 @@
         "transform": {
           "$ref": "#/$defs/Template"
         },
-        "filter": {
+        "Filter": {
```
yashmehrotra, line 615: Should `Filter` and `Auth` be capitalised here ?

Issue #947, opened 2023-04-19
Component spec is missing labels


adityathebe:

adityathebe:
@moshloop Should it be named `labels or` `tags`?

moshloop:
Use labels

PR #1112, opened 2023-07-03
add labels to component spec

adityathebe:
Resolves: https://github.com/flanksource/canary-checker/issues/947


moshloop:
@adityathebe don't we need changes in duty to persist the labels ?

adityathebe:
> @adityathebe don't we need changes in duty to persist the labels ?

We are currently not using component model from duty. We're using `pkg/system_api.go`

PR #1118, opened 2023-07-04
changes to the graph API

adityathebe:


pkg/api/api.go
```diff
@@ -20,12 +23,46 @@ type Response struct {
...
-type DetailResponse struct {
+
+// Latency metrics in milliseconds
+type Latency struct {
+	P95 int `json:"p95,omitempty"`
+	P97 int `json:"p97,omitempty"`
+	P99 int `json:"p99,omitempty"`
+}
+
+func measureLatency(data []pkg.Timeseries) Latency {
```
moshloop, line 34: We need todo in the DB,   If there are 1000 statuses i am expecting 1000 passed and latency measurements on those 1000, and then the graph to show an aggregated view

pkg/api/api.go
```diff
@@ -41,18 +78,65 @@ func CheckDetails(c echo.Context) error {
...
+	for _, t := range ts {
+		sum += t.Duration
+	}
+
+	output.Duration = int(math.Ceil(float64(sum) / float64(len(ts))))
+	return output
+}
+
+func aggregateWindow(ts []pkg.Timeseries, maxDataPoints int) []pkg.Timeseries {
+	if len(ts) <= maxDataPoints {
```
moshloop, line 113: This needs to be at the db layer.

pkg/api/api.go
adityathebe: The latency is measured before aggregation so it's already like that.
Unless you mean to measure the latency in the SQL query itself?

pkg/api/api.go
adityathebe: By that do you mean in the `cache` package or in the SQL itself?

pkg/api/api.go
```diff
@@ -87,3 +129,37 @@ func HealthSummary(c echo.Context) error {
 	}
 	return c.JSON(http.StatusOK, apiResponse)
 }
+
+func getBestPartitioner(totalChecks int, rangeDuration time.Duration) time.Duration {
+	if totalChecks <= desiredNumOfCheckStatuses {
```
adityathebe, line 134: > if I have a check that runs once per hour, then I can go at least 4d before needing to aggregate

@moshloop In this case you mentioned earlier, we would have still not performed aggregation because the total checks (96) would be less than the desired num of checks (100).

However, I have made improvements to handle cases where aggregation is not done even if the total number of checks are above 100 _(example: an hourly check for the duration of 6 days, 8th case in the unit test)_

PR #1136, opened 2023-07-10
chore: delete transformed checks from cron

yashmehrotra:
Fixes: https://mission-control.azure.lab.flanksource.com/incidents/018929f8-be24-2368-7983-a4e47955d5b5

### DIFF
--- a/pkg/db/canary.go
+++ b/pkg/db/canary.go
@@ -96,10 +96,6 @@ func PersistCheck(check pkg.Check, canaryID uuid.UUID) (uuid.UUID, error) {
 		"deleted_at":  nil,
 	}
 
-	// Deletions for transformed checks are handled separately
-	if check.Transformed {
-		delete(assignments, "deleted_at")
-	}
 	tx := Gorm.Clauses(clause.OnConflict{
 		Columns:   []clause.Column{{Name: "canary_id"}, {Name: "type"}, {Name: "name"}, {Name: "agent_id"}},
 		DoUpdates: clause.Assignments(assignments),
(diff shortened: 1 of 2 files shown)


pkg/jobs/canary/canary_jobs.go
```diff
@@ -51,7 +51,8 @@ func StartScanCanaryConfigs(dataFile string, configFiles []string) {
 type CanaryJob struct {
 	*kommons.Client
 	Kubernetes kubernetes.Interface
-	v1.Canary
+	Canary     v1.Canary
+	DBCanary   pkg.Canary
```
yashmehrotra, line 55: Will use DB object from now on, will slowly deprecate uses of the old spec

pkg/jobs/canary/canary_jobs.go
```diff
@@ -111,6 +112,11 @@ func (job CanaryJob) Run() {
 		logger.Errorf("error deleting transformed checks for canary %s: %v", job.Canary.GetPersistedID(), err)
 	}
 
+	if len(checksToMarkHealthy) > 0 {
+		// Remove from cron and let the transformed checks repopulate
```
moshloop, line 116: Not sure I understand this- Why do we remove the schedule?

yashmehrotra: When a transformed check is marked healthy, it is technically soft deleted but it still remains in the cronjob. We delete the cronentry and the SyncCanaryJob repopulates the cron with non deleted checks

pkg/db/canary.go
```diff
@@ -96,10 +96,6 @@ func PersistCheck(check pkg.Check, canaryID uuid.UUID) (uuid.UUID, error) {
 		"deleted_at":  nil,
 	}
 
-	// Deletions for transformed checks are handled separately
```
yashmehrotra, line 99: @moshloop 
Not sure why we put this, removing it

Issue #1146, opened 2023-07-12
Option to disable 1 or more check types


moshloop:

adityathebe:
@moshloop so we're creating a new table _(disabled_check_types)_ for this and then checking it before scheduling?

PR #1156, opened 2023-07-17
feat: disabled checks

adityathebe:
Resolves: https://github.com/flanksource/canary-checker/issues/1146

### DIFF
--- a/checks/runchecks.go
+++ b/checks/runchecks.go
@@ -26,7 +26,7 @@ func getDisabledChecks(ctx *context.Context) (map[string]struct{}, error) {
 		return result, nil
 	}
 
-	rows, err := ctx.DB().Raw("SELECT name FROM properties WHERE name LIKE 'check.disabled.%'").Rows()
+	rows, err := ctx.DB().Raw("SELECT name FROM properties WHERE name LIKE 'check.disabled.%' AND value = 'true'").Rows()
 	if err != nil {
 		return nil, err
 	}
--- a/checks/runchecks.go
+++ b/checks/runchecks.go
@@ -36,6 +36,10 @@ func getDisabledChecks() (map[string]struct{}, error) {
 		result[name] = struct{}{}
 	}
 
+	if rows.Err() != nil {
+		return nil, rows.Err()
+	}
+
 	disabledChecks = result
 	return disabledChecks, nil
 }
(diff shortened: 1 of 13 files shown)


checks/runchecks.go
```diff
@@ -11,15 +12,56 @@ import (
...
-func RunChecks(ctx *context.Context) []*pkg.CheckResult {
+// A list of check types that are permanently disabled.
+var disabledChecks map[string]struct{}
+
+func getDisabledChecks() (map[string]struct{}, error) {
+	if disabledChecks != nil {
+		return disabledChecks, nil
+	}
+
+	rows, err := db.Gorm.Raw("SELECT name FROM disabled_checks").Rows()
```
moshloop, line 23: This will throw an error for `canary-checker run` - Could we add an e2e test so that it picks it up in future ?

checks/runchecks.go
```diff
@@ -11,15 +12,60 @@ import (
...
+	if disabledChecks != nil {
+		return disabledChecks, nil
+	}
+
+	result := make(map[string]struct{})
+	if ctx.DB() == nil {
+		return result, nil
+	}
+
+	rows, err := ctx.DB().Raw("SELECT value FROM properties WHERE name = 'check'").Rows()
```
moshloop, line 28: I think the format should be: name=check.<type>.disabled, value = true 
We might also want to populate the properties table from a `canary-checker.properties`  file

cmd/serve.go
```diff
@@ -47,6 +51,35 @@ var Serve = &cobra.Command{
 	},
 }
 
+func loadPropertiesToDB(db *gorm.DB, filename string) error {
+	if err := db.Exec("DELETE FROM properties WHERE name LIKE 'check.disabled.%'").Error; err != nil {
```
moshloop, line 55: I don't think we should be doing this, rather upsert the properties

checks/runchecks.go
```diff
@@ -11,15 +13,60 @@ import (
 	"github.com/flanksource/commons/logger"
 )
 
-func RunChecks(ctx *context.Context) []*pkg.CheckResult {
+// A list of check types that are permanently disabled.
+var disabledChecks map[string]struct{}
+
+func getDisabledChecks(ctx *context.Context) (map[string]struct{}, error) {
+	if disabledChecks != nil {
```
moshloop, line 20: I might make sense to use a  5 minute cache here

adityathebe: @moshloop don't we just need to read this once (on startup)?

moshloop: The startup is more for licensing / bootstraping,  If one particular check type is causing issues in production, then we might want to disable it via update to properties

PR #1167, opened 2023-07-20
fix: respect agent_id and deleted_at for topologies

yashmehrotra:


pkg/db/topology.go
```diff
@@ -63,7 +63,11 @@ func GetTopology(ctx context.Context, id string) (*v1.Topology, error) {
 func GetAllTopologiesForSync() ([]v1.Topology, error) {
 	var v1topologies []v1.Topology
 	var topologies []pkg.Topology
-	if err := Gorm.Table("topologies").Find(&topologies).Where("deleted_at is NULL AND namespace != 'push'").Error; err != nil {
```
yashmehrotra, line 66: Since `.Find` was being called first and is a finisher method, we were never respecting deleted at

PR #1169, opened 2023-07-24
fix: add canary_id to job_history and cleanup sync canary function

yashmehrotra:
Fixes: https://github.com/flanksource/duty/issues/229

@moshloop @adityathebe 
Removed some dead logic and redundant conditions which I don't think was necessary. Please do a double check on this

### DIFF
--- a/pkg/jobs/canary/canary_jobs.go
+++ b/pkg/jobs/canary/canary_jobs.go
@@ -276,8 +276,13 @@ func ScanCanaryConfigs() {
 var canaryUpdateTimeCache = sync.Map{}
 
 // TODO: Refactor to use database object instead of kubernetes
-func SyncCanaryJob(canary v1.Canary) error {
-	if !canary.DeletionTimestamp.IsZero() || canary.Spec.GetSchedule() == "@never" {
+func SyncCanaryJob(dbCanary pkg.Canary) error {
+	canary, err := dbCanary.ToV1()
+	if err != nil {
+		return err
+	}
+
+	if canary.Spec.GetSchedule() == "@never" {
 		DeleteCanaryJob(canary.GetPersistedID())
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


pkg/jobs/canary/canary_jobs.go
```diff
@@ -333,52 +334,25 @@ func SyncCanaryJob(canary v1.Canary) error {
...
 		return
 	}
 
 	for _, c := range canaries {
-		canary, err := c.ToV1()
-		if err != nil {
-			logger.Errorf("Error parsing canary[%s]: %v", c.ID, err)
-			jobHistory.AddError(err.Error())
-			continue
-		}
```
yashmehrotra, line 353: Not required, we are doing this in the SyncCanary func

pkg/jobs/canary/canary_jobs.go
```diff
@@ -333,52 +334,25 @@ func SyncCanaryJob(canary v1.Canary) error {
...
-
-		if len(canary.Status.Checks) == 0 && len(canary.Spec.GetAllChecks()) > 0 {
-			logger.Infof("Persisting %s as it has no checks", canary.Name)
-			pkgCanary, _, _, err := db.PersistCanary(*canary, canary.Annotations["source"])
-			if err != nil {
-				logger.Errorf("Failed to persist canary %s: %v", canary.Name, err)
-				jobHistory.AddError(err.Error())
-				continue
-			}
```
yashmehrotra, line 362: Since these canaries are always fetched from database, no persistence logic is needed

pkg/jobs/canary/canary_jobs.go
```diff
@@ -333,52 +334,25 @@ func SyncCanaryJob(canary v1.Canary) error {
...
-
-			v1canary, err := pkgCanary.ToV1()
-			if err != nil {
-				logger.Errorf("Failed to convert canary to V1 %s: %v", canary.Name, err)
-				jobHistory.AddError(err.Error())
-				continue
-			}
```
yashmehrotra, line 369: redundant conversion

pkg/jobs/canary/canary_jobs.go
adityathebe: @yashmehrotra we were also saving checks from the canary spec in here. Now, checks aren't being saved.

adityathebe: @yashmehrotra I'm not talking about component checks. If you create a canary from the UI then the checks are never saved.

PR #1174, opened 2023-07-26
fix: sync checks for canaries added from ui

yashmehrotra:


pkg/jobs/canary/canary_jobs.go
```diff
@@ -415,7 +415,7 @@ func ReconcileDeletedCanaryChecks() {
 		ID        string
```
adityathebe, line 415: @yashmehrotra, do we still need this job since we have this now
https://github.com/flanksource/duty/pull/206/files

pkg/jobs/canary/canary_jobs.go
```diff
@@ -365,6 +365,48 @@ func DeleteCanaryJob(id string) {
 	CanaryScheduler.Remove(entry.ID)
 }
 
+func ReconcileCanaryChecks() {
```
adityathebe, line 368: missing logging of errors to the job history

PR #1188, opened 2023-07-31
refactor: move common test logic to separate package

yashmehrotra:
Update go.mod after https://github.com/flanksource/duty/pull/235 is merged


yashmehrotra:
@moshloop @adityathebe Should I move this to the commons repo ? we can probably re-use this for incident commander as well

moshloop:
> @moshloop @adityathebe Should I move this to the commons repo ? we can probably re-use this for incident commander as well

I don't think commons is the right place as it will incresae the dependency tree, duty is probably a better place

Issue #1137, opened 2023-07-11
AWS / GCP / Azure CLI checks


moshloop:
i.e. use connections to run CLI commands

```yaml
exec:
 - connections
       aws: 
           connection: connection://aws/profile
       kubectl:
          connection: ...
   exec:  aws sts get-caller identity
```

Update both the minimal and full docker containers to install via https://github.com/alexellis/arkade: kubectl, stern, yq, jq, aws, azure, gcp CLI's

PR #1212, opened 2023-08-07
feat: AWS / GCP / Azure CLI checks

adityathebe:
Resolves: https://github.com/flanksource/canary-checker/issues/1137


api/v1/checks.go
```diff
@@ -822,6 +822,7 @@ type AWSConnection struct {
 	ConnectionName string       `yaml:"connection,omitempty" json:"connection,omitempty"`
 	AccessKey      types.EnvVar `yaml:"accessKey" json:"accessKey,omitempty"`
 	SecretKey      types.EnvVar `yaml:"secretKey" json:"secretKey,omitempty"`
+	SessionToken   types.EnvVar `yaml:"sessionToken,omitempty" json:"sessionToken,omitempty"`
```
moshloop, line 825: I don't think a user would ever specify this as they expire

build/full/Dockerfile
```diff
@@ -51,42 +51,62 @@ RUN curl -L https://dlcdn.apache.org//jmeter/binaries/apache-jmeter-${JMETER_VER
...
+# Install CLIs with arkade as canary user
+RUN arkade get kubectl
+RUN arkade get stern
+RUN arkade get jq
+RUN arkade get yq
+ENV PATH="${PATH}:/var/lib/canary/.arkade/bin/"
+ENV PATH="${PATH}:/var/lib/canary/bin/"
+
+# Install AWS CLI
+RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
```
moshloop, line 103: Do this in 1 command and then remove the zip file to save space in the docker image - same applies to the others

checks/exec.go
```diff
@@ -49,14 +54,87 @@ func execPowershell(check v1.ExecCheck, ctx *context.Context) pkg.Results {
...
-	cmd := osExec.Command("bash", "-c", *check.Script)
+	fields := strings.Fields(check.Script)
+	if len(fields) == 0 {
+		return []*pkg.CheckResult{result.Failf("no script provided")}
+	}
+
+	cmd := osExec.Command("bash", "-c", check.Script)
+
+	// Setup connection details
+	switch strings.ToLower(fields[0]) {
```
moshloop, line 72: You can specify multiple connections - e.g. aws and azure

checks/exec.go
```diff
@@ -49,14 +54,87 @@ func execPowershell(check v1.ExecCheck, ctx *context.Context) pkg.Results {
...
+		}
+
+		cmd.Env = os.Environ()
+		cmd.Env = append(cmd.Env, "AWS_EC2_METADATA_DISABLED=true") // https://github.com/aws/aws-cli/issues/5262#issuecomment-705832151
+		cmd.Env = append(cmd.Env, fmt.Sprintf("AWS_SHARED_CREDENTIALS_FILE=%s", configPath))
+		if check.Connections.AWS.Region != "" {
+			cmd.Env = append(cmd.Env, fmt.Sprintf("AWS_DEFAULT_REGION=%s", check.Connections.AWS.Region))
+		}
+
+	case "az":
```
moshloop, line 95: I think use `azure`

checks/exec.go
```diff
@@ -73,8 +151,46 @@ func runCmd(cmd *osExec.Cmd, result *pkg.CheckResult) (results pkg.Results) {
...
-		return result.Failf("non-zero exit-code: %d: %s %s", details.ExitCode, details.Stdout, details.Stderr).ToSlice()
+		return result.Failf("non-zero exit-code: %d. (stdout=%s) (stderr=%s)", details.ExitCode, details.Stdout, details.Stderr).ToSlice()
 	}
+
 	results = append(results, result)
 	return results
 }
+
+func saveConfig(dirPrefix string, configTemplate *textTemplate.Template, view any) (string, error) {
+	dirPath, err := os.MkdirTemp("", dirPrefix)
```
moshloop, line 162: The credentials should not be in the /tmp folder which is world readable, do it in the current directory

checks/exec.go
adityathebe: I think `az` would be the standard way to use the azure cli. I have azure cli installed locally, and I can only access it with `az`. Unless we setup `alias azure='az'`

api/v1/checks.go
adityathebe: It was only needed to make it work with the aws-credentials github action.

checks/exec.go
moshloop: `az` maybe the command name, but it isn't clear it means azure.

PR #1222, opened 2023-08-17
feat: custom metrics for check results

yashmehrotra:


checks/metrics.go
```diff
@@ -0,0 +1,125 @@
...
+			logger.Errorf("Invalid type for check.metrics %s for check[%s]", metricsSpec.Type, c.GetName())
+			return
+		}
+	}
+
+	tplValue := v1.Template{Template: metricsSpec.Value}
+	for _, r := range results {
+		templateInput := map[string]any{
+			"result": r.Data,
+			"check": map[string]any{
```
moshloop, line 86: Probably need a duration in here as well

fixtures/minimal/http_pass_single.yaml
```diff
@@ -11,6 +11,14 @@ spec:
       name: http-deprecated-endpoint
     - name: http-minimal-check
       url: https://httpbin.demo.aws.flanksource.com/status/200
+      metrics:
```
moshloop, line 14: can we also add examples for junit e.g. number of tests, passing, failing, time

checks/metrics.go
```diff
@@ -0,0 +1,125 @@
...
+				"name":        c.GetName(),
+				"description": c.GetDescription(),
+				"labels":      c.GetLabels(),
+				"endpoint":    c.GetEndpoint(),
+			},
+		}
+		valRaw, err := template(ctx.New(templateInput), tplValue)
+		if err != nil {
+			logger.Errorf("Error templating value for check.metrics template %s for check[%s]", metricsSpec.Value, c.GetName())
```
adityathebe, line 95: need to log the error as well. And should we proceed with execution after an error? I think we need a continue

fixtures/minimal/http_pass_single.yaml
yashmehrotra: Added

api/external/metrics.go
```diff
@@ -0,0 +1,9 @@
+package external
+
+// +kubebuilder:object:generate=true
+type Metrics struct {
```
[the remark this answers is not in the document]
yashmehrotra: Fixed

checks/metrics.go
yashmehrotra: Added

PR #1234, opened 2023-08-24
chore: update custom metric label type

yashmehrotra:


moshloop:
@yashmehrotra can you fix the linting error and possibly add a test as well ?

PR #1235, opened 2023-08-24
feat: use cel for kubernetes bundle

yashmehrotra:
Depends on https://github.com/flanksource/gomplate/pull/19


moshloop:
@yashmehrotra  could we add a test for this ?

Issue #1232, opened 2023-08-23
[bug] Canary-check: Pod: Issue with scheduling pods in case taints and tolerations using


msergg:
In case creating pod check in example:

```
apiVersion: canaries.flanksource.com/v1
kind: Canary
metadata:
  name: pod-check
spec:
  interval: 30
  pod:
    - name: golang
      spec: |
        apiVersion: v1
        kind: Pod
        metadata:
          name: hello-world-golang
          namespace: default
          labels:
            app: hello-world-golang
        spec:
          containers:
            - name: hello
              image: quay.io/toni0/hello-webserver-golang:latest
      port: 8080
      path: /foo/bar
      scheduleTimeout: 20000
      readyTimeout: 10000
      httpTimeout: 7000
      deleteTimeout: 12000
      ingressTimeout: 10000
      deadline: 60000
      httpRetryInterval: 200
      expectedContent: bar
      expectedHttpStatuses: [200, 201, 202]
      priorityClass: canary-checker-priority
```

And using some configuration for node taints and toleration, 
we getting error, do not have possibility to schedule pod.

It is caused due to using logic configured in pod check
with next node logic as node selector in previous steps:
```
	pod.Spec.NodeSelector = map[string]string{
		"kubernetes.io/hostname": nodeName,
	}
```
So kube scheduler could not schedule pod to free untained instance

https://github.com/flanksource/canary-checker/blob/cc1c86306aa5013e670add2da43136bf6adf0eee/checks/pod.go#L92

Please remove this part or make it optional and configurable
```
	pod.Spec.NodeSelector = map[string]string{
		"kubernetes.io/hostname": nodeName,
	}
```

PR #1239, opened 2023-08-28
[bug] Canary-check: Pod: Issue with scheduling pods in case taints and tolerations using #1232

msergg:
https://github.com/flanksource/canary-checker/issues/1232

### DIFF
--- a/checks/pod.go
+++ b/checks/pod.go
@@ -19,9 +19,11 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 
 	canaryv1 "github.com/flanksource/canary-checker/api/v1"
+
 	"github.com/flanksource/canary-checker/pkg"
 	"github.com/flanksource/canary-checker/pkg/metrics"
 	"github.com/flanksource/commons/logger"
+
 	perrors "github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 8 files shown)


moshloop:
Hi @msergg and thank you for the contribution 

The reasoning behind this behaviour is to test the schedulability of every node in a cluster - mostly relevant for on-premise / bare-metal clusters - however I do agree that it is perhaps not the best default behaviour, could you gate the old behaviour in a `roundRobinNodes` option?

msergg:
> roundRobinNodes

Do you mean that there is an option in documentation for roundRobinNodes? Or need to develop this option?


About default behaviour, a also got errors during scheduling on both premise and cloud clusters,
in case we have some topology constraints or worse - some selected nodes could be already full and scheduler could not deploy test pod to this insance - so check will fail and it not show correct cluster scheduling state


I thing better to remove such behaviour or configure it to seperate option if some one need it

moshloop:
@msergg - The `roundRobinNodes` option doesn't exist, I am suggesting that you add it as part of your PR

msergg:
@moshloop 
I have added an option usePreparedNodeSheduling
how do you test  it locally if shema changes? do you have any procedure?

moshloop:
@msergg If you run `make resources` it will regenerate all the CRD's and manifest files

msergg:
@moshloop pls check changes , seem this logic should be enough

checks/pod.go
```diff
@@ -89,9 +91,13 @@ func (c *PodChecker) newPod(podCheck canaryv1.PodCheck, nodeName string) (*v1.Po
 	pod.Labels[nameLabel] = pod.Name
 	pod.Labels[podCheckSelector] = c.podCheckSelectorValue(podCheck)
 	pod.Labels[podGeneralSelector] = "true"
-	pod.Spec.NodeSelector = map[string]string{
-		"kubernetes.io/hostname": nodeName,
+
+	if podCheck.RoundRobinNodes == false {
```
moshloop, line 95:
```suggestion
	if podCheck.RoundRobinNodes  {
```

msergg:
cannot pass all license/cla , some commits were pushed from another unexisting users

msergg:
@moshloop could you pls approve next steps or we need to fix smths?

moshloop:
@msergg If you squash all your changes under `msergg` and not `serhii.martyniuk` the CLA check will pass

msergg:
@moshloop could you pls review?

PR #1246, opened 2023-09-02
fix: update on conflict clause for canaries

yashmehrotra:


pkg/controllers/canary_controller.go
```diff
@@ -95,13 +95,13 @@ func (r *CanaryReconciler) Reconcile(ctx gocontext.Context, req ctrl.Request) (c
...
 
-	dbCanary, checks, changed, err := db.PersistCanary(*canary, "kubernetes/"+canary.GetPersistedID())
+	dbCanary, checks, err := db.PersistCanary(*canary, "kubernetes/"+canary.GetPersistedID())
 	if err != nil {
 		return ctrl.Result{Requeue: true}, err
 	}
 
 	// Sync jobs if canary is created or updated
-	if canary.Generation == 1 || changed {
+	if canary.Generation == 1 {
```
yashmehrotra, line 104: @moshloop This was previously being called all the time since the `INSERT ON CONFLICT` query in `PersistCanary` func always returned `changed` as `true`

I've removed the `changed` var, should this sync be called always ?

Right now, there is a case where if the user updates the CRD and it will be out of sync till the automated sync schedule is run (2 minute interval period)

The reconcile function is called a lot because even the status updates to canaries are considered updates.

The `SyncCanaryJob` is idempotent, so should I just remove the `generation` check and call this everytime ?

yashmehrotra: We can also have a small cache where we keep the spec field and compare that to call the Persist and Sync funcs

moshloop: Yeh, I think we should add a cache thats updated after persistence, and don't even touch the db if nothing in spec/metadata has changed. 

I am also not sure why we have a status.ObserverdGeneration if we are not reading it anywhere.

Issue #1255, opened 2023-09-07
Checks are never created for Canaries created from the UI


adityathebe:
I created a new HTTP check from the UI however only the canary is saved. The corresponding check is never saved.

PR #1256, opened 2023-09-11
fix: canaries created from UI were not being saved

adityathebe:
Resolves: https://github.com/flanksource/canary-checker/issues/1255

### DIFF
--- a/pkg/cache/postgres.go
+++ b/pkg/cache/postgres.go
@@ -54,7 +54,7 @@ func (c *postgresCache) AddCheckFromStatus(check pkg.Check, status pkg.CheckStat
 		return uuid.Nil, nil
 	}
 
-	if !check.Transformed {
+	if check.ID != uuid.Nil {
 		return uuid.Nil, nil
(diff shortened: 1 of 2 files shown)


pkg/cache/postgres.go
```diff
@@ -54,7 +54,7 @@ func (c *postgresCache) AddCheckFromStatus(check pkg.Check, status pkg.CheckStat
 		return uuid.Nil, nil
 	}
 
-	if !check.Transformed {
+	if check.ID != uuid.Nil {
```
moshloop, line 57: @yashmehrotra How will this affect the transformed canaries ?

yashmehrotra: It will work fine, since transformed checks won't have the check id set

@adityathebe Can you confirm by running the alertmanager check

adityathebe: @moshloop @yashmehrotra can confirm that alertmanager checks were created.

Issue #1289, opened 2023-10-04
CRD status for http checks is always failed


moshloop:

moshloop:
+ they seem to transition on every check on the CRD,  but on the DB they don't

PR #1295, opened 2023-10-05
fix: panic err in counter metric creation

yashmehrotra:
Fixes: https://github.com/flanksource/canary-checker/issues/1289


checks/metrics.go
```diff
@@ -40,9 +41,20 @@ func getOrAddPrometheusMetric(name, metricType string, labelNames []string) (pro
 	return collector, prometheus.Register(collector)
 }
 
-func getWithEnvironment(ctx *context.Context, r *pkg.CheckResult) *context.Context {
+func getWithEnvironment(ctx *context.Context, r *pkg.CheckResult) (*context.Context, error) {
+	// Convert result Data into JSON for templating
+	// as gomplate does not render child map data correctly
+	var rData map[string]any
```
moshloop, line 47: Fixing this issue in https://github.com/flanksource/canary-checker/pull/1288 
I want to try and avoid unnecessary marshalling  - if it needs to happen should be in gomplate.

pkg/metrics/metrics.go
```diff
@@ -207,18 +210,17 @@ func Record(canary v1.Canary, result *pkg.CheckResult) (_uptime pkg.Uptime, _lat
...
-				getOrCreateGauge(m)
 				if err := getOrCreateGauge(m); err != nil {
-					result.ErrorMessage(fmt.Errorf("cannot create gauge %s with labels %v", m.Name, m.Labels))
+					logger.Errorf("cannot create gauge %s with labels %v: %w", m.Name, m.Labels, err)
 				}
 
 			case HistogramType:
 				if err := getOrCreateHistogram(m); err != nil {
-					result.ErrorMessage(fmt.Errorf("cannot create histogram %s with labels %v", m.Name, m.Labels))
+					logger.Errorf("cannot create histogram %s with labels %v: %w", m.Name, m.Labels, err)
```
moshloop, line 223: Why would failing to create a metric, not fail the check ?

yashmehrotra: If the check passes but if there is an instrumentation error from our side, the check in itself still passes right ?

It might create false positives this way

PR #1297, opened 2023-10-06
refactor: component property lookup flow

yashmehrotra:


pkg/topology/run.go
```diff
@@ -81,7 +81,16 @@ func forEachComponent(ctx *ComponentContext, spec *v1.ComponentSpec, component *
...
 
-		for _, p := range props {
+		// TODO: Ask Moshe Can for each handle component list
+		if err := mergeComponentProperties(pkg.Components{component}, props); err != nil {
+			continue
+		}
+
+		var properties pkg.Properties
+		if err := json.Unmarshal(props, &properties); err != nil {
+			return err
```
yashmehrotra, line 91: Is this the correct way to handle properties for `forEach` ?

pkg/topology/run_test.go
```diff
@@ -0,0 +1,64 @@
...
+	ginkgo.It("should create correct component properties", func() {
+
+		rootComponent := Run(opts, topology)
+		Expect(len(rootComponent[0].Components)).To(Equal(3))
+
+		componentA := rootComponent[0].Components[0]
+		componentB := rootComponent[0].Components[1]
+		componentC := rootComponent[0].Components[2]
+
+		Expect(string(componentA.Properties.AsJSON())).To(MatchJSON(`[{"name":"error_percentage","value":1,"max":100},{"name":"owner","text":"team-a"},{"name":"company","text":"Acme"},{"name":"location","text":"Mars"}]`))
```
moshloop, line 60: Is this not missing `min: 0`

yashmehrotra: It was due to min field being 0 value, good that we caught it in the test

Have made that field a pointer

PR #1370, opened 2023-10-23
feat: new job type for sync canary jobs

yashmehrotra:


yashmehrotra:
image

pkg/jobs/canary/canary_jobs.go
```diff
@@ -50,36 +56,36 @@ func StartScanCanaryConfigs(dataFile string, configFiles []string) {
...
 }
 
 var minimumTimeBetweenCanaryRuns = 10 * time.Second
 var canaryLastRuntimes = sync.Map{}
 
-func (job CanaryJob) Run() {
-	if runner.IsCanaryIgnored(&job.Canary.ObjectMeta) {
-		return
+func (j CanaryJob) Run(ctx dutyjob.JobRuntime) error {
+	ctx.GetSpan().SetAttributes(attribute.String("canary-id", j.DBCanary.ID.String()))
```
moshloop, line 73: I don't think we should be recording traces for canary runs - they are traced internally using check_status, not sure what pushing them to opentelemetry would add ? A slow canary is more likely due to a slow remote server than a slow canary checker db ?

yashmehrotra: Dit it for general observability, like record all transactions to have complete overview and have a general idea of how much time the checks run is taking compared to the other database calls in the job

pkg/db/canary.go
```diff
@@ -278,7 +279,7 @@ func FindCheck(canary pkg.Canary, name string) (*pkg.Check, error) {
 	return &model, nil
 }
 
-func FindDeletedChecksSince(ctx context.Context, since time.Time) ([]string, error) {
+func FindDeletedChecksSince(ctx gocontext.Context, since time.Time) ([]string, error) {
```
adityathebe, line 282: Maybe use duty's context for this as well

yashmehrotra: Done 👍🏽

PR #1380, opened 2023-10-26
chore: change es fixture secret name

yashmehrotra:


yashmehrotra:
@moshloop Kustomize build was failing on the cluster because of resources being named the same

PR #1383, opened 2023-10-27
fix: config lookup for properties with test

yashmehrotra:


pkg/topology/run.go
```diff
@@ -264,7 +264,7 @@ func lookupConfig(ctx *ComponentContext, property *v1.Property) (*pkg.Property,
 	}
 
 	templateEnv := map[string]any{
-		"config": _config.Spec,
+		"config": *_config.Spec,
```
yashmehrotra, line 267: @moshloop The javascript VM was breaking since we were passing a pointer, using value fixed the problem

Should we add a check in gomplate to throw error when it gets a pointer ?

moshloop: Or just deref it automatically ?

PR #1386, opened 2023-10-28
chore: use default prometheus url for checks

yashmehrotra:


cmd/root.go
```diff
@@ -61,7 +61,7 @@ func ServerFlags(flags *pflag.FlagSet) {
 	flags.StringSliceVar(&pushServers, "push-servers", []string{}, "push check results to multiple canary servers")
 	flags.StringSliceVar(&pullServers, "pull-servers", []string{}, "push check results to multiple canary servers")
 	flags.StringVar(&runner.RunnerName, "name", "local", "Server name shown in aggregate dashboard")
-	flags.StringVar(&prometheusURL, "prometheus", "", "URL of the prometheus server that is scraping this instance")
+	flags.StringVar(&prometheus.PrometheusURL, "prometheus", "", "URL of the prometheus server that is scraping this instance")
```
moshloop, line 64: Can we add this as a chart value as well ?

Issue #1364, opened 2023-10-19
Webhook Check


moshloop:
```yaml
webhook:
   - name: # globally unique name for the webhook
      token: 
          name: 'token'. # validate this token from query params or headers
          valueFrom: ...
```

Create an endpoint `/webhook/{check_id}|{name}` if more than 1 matching webhook return an error

Normal transform functions are then used to extract metrics, or check details with the template vars:

- headers
- content
- json

Create an example of alertmanagers https://prometheus.io/docs/alerting/latest/configuration/#webhook_config

PR #1387, opened 2023-10-30
feat: webhook check

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/1364


api/v1/canary_types.go
```diff
@@ -73,6 +73,7 @@ type CanarySpec struct {
 	AlertManager   []AlertManagerCheck   `yaml:"alertmanager,omitempty" json:"alertmanager,omitempty"`
 	Dynatrace      []DynatraceCheck      `yaml:"dynatrace,omitempty" json:"dynatrace,omitempty"`
 	AzureDevops    []AzureDevopsCheck    `yaml:"azureDevops,omitempty" json:"azureDevops,omitempty"`
+	Webhook        []WebhookCheck        `yaml:"webhook,omitempty" json:"webhook,omitempty"`
```
moshloop, line 76: I don't think this needs to be an array

api/v1/canary_types.go
```diff
@@ -197,6 +198,9 @@ func (spec CanarySpec) GetAllChecks() []external.Check {
 	for _, check := range spec.Opensearch {
 		checks = append(checks, check)
 	}
+	for _, check := range spec.Webhook {
```
moshloop, line 201: Its probably not necessary to include this in GetAllChecks

adityathebe: I had it excluded initially but the check was never saved by the `SyncCanaryJob` because `RunChecks` returned 0 results.

checks/runchecks.go
```diff
@@ -184,3 +188,55 @@ func processTemplates(ctx *context.Context, r *pkg.CheckResult) *pkg.CheckResult
...
+		}
+
+		transformedChecksAdded := cache.PostgresCache.Add(pkg.FromV1(result.Canary, result.Check), pkg.FromResult(*result))
+		transformedChecksCreated = append(transformedChecksCreated, transformedChecksAdded...)
+		for _, checkID := range transformedChecksAdded {
+			checkIDDeleteStrategyMap[checkID] = result.Check.GetTransformDeleteStrategy()
+		}
+	}
+
+	checkDeleteStrategyGroup := make(map[string][]string)
```
moshloop, line 216: I don't think we can reuse the transformation deletion logic here, With a webhook the existingTransformed checks are not relevant - We probably need to add a deleted_at value in TransformedCheckResult and in the transformaton function set deleted_at e.g. when an alert status is resolved.

api/v1/canary_types.go
```diff
@@ -74,6 +74,7 @@ type CanarySpec struct {
 	AlertManager   []AlertManagerCheck   `yaml:"alertmanager,omitempty" json:"alertmanager,omitempty"`
 	Dynatrace      []DynatraceCheck      `yaml:"dynatrace,omitempty" json:"dynatrace,omitempty"`
 	AzureDevops    []AzureDevopsCheck    `yaml:"azureDevops,omitempty" json:"azureDevops,omitempty"`
+	Webhook        *WebhookCheck         `yaml:"webhook,omitempty" json:"webhook,omitempty"`
```
yashmehrotra, line 77: Why is this not a list ?

Issue #1423, opened 2023-11-06
Webhook transformations delete all old checks


moshloop:
Webhooks should not use the transformation strategies for deleting and/or updating check status based on abscence e.g. transformed should be false.

Checks created by webhooks can only be deleted by a subsequent webhook that sets a deleted_at on the check

it might make senses to use a mapping field instead of a transform field . 

+ add tests for sending multiple payloads to a webhook (creating both new checks, updating existing checks and removing some)

PR #1429, opened 2023-11-07
fix: ignore webhook checks when using delete transformation strategy

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/1423

### DIFF
--- a/pkg/api/webhook_test.go
+++ b/pkg/api/webhook_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/flanksource/canary-checker/checks"
 	"github.com/flanksource/canary-checker/pkg/db"
 	canaryJobs "github.com/flanksource/canary-checker/pkg/jobs/canary"
+	"github.com/flanksource/canary-checker/pkg/utils"
 	"github.com/flanksource/commons/http"
 	"github.com/flanksource/commons/logger"
 	"github.com/flanksource/duty/job"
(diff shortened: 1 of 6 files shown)


pkg/jobs/canary/canary_jobs.go
```diff
@@ -117,8 +117,14 @@ func (j CanaryJob) Run(ctx dutyjob.JobRuntime) error {
 		results = append(results, pkg.Success(canaryCtx.Canary.Spec.Webhook, canaryCtx.Canary))
 	}
 
-	// Get transformed checks before and after, and then delete the olds ones that are not in new set
-	existingTransformedChecks, _ := db.GetTransformedCheckIDs(ctx.Context, canaryID)
+	// Get transformed checks before and after, and then delete the olds ones that are not in new set.
+	// NOTE: Webhook checks, although they are transformed, are handled entirely by the webhook caller
+	// and should not be deleted manually in here.
+	existingTransformedChecks, err := db.GetTransformedCheckIDs(ctx.Context, canaryID, checks.WebhookCheckType)
```
adityathebe, line 123: I think this should suffice

pkg/api/webhook_test.go
```diff
@@ -191,4 +225,77 @@ var _ = ginkgo.Describe("Test Sync Canary Job", ginkgo.Ordered, func() {
 		Expect(err).To(BeNil())
 		Expect(result.DeletedAt).To(Not(BeNil()))
 	})
+
+	ginkgo.It("Should create 1 new check from the webhook and deleted one", func() {
+		body := `
```
moshloop, line 230: We need to split this into multiple individual calls to test how the webhook behaves after multiple subsequent calls with data that differs from the first. 
3 x firing
1 x resolved

and then verify 2 exist and 1 is deleted

PR #1453, opened 2023-11-20
feat: artifacts

adityathebe:


checks/connection_filesystem.go
```diff
@@ -0,0 +1,367 @@
...
+	if err := writer.Close(); err != nil {
+		return nil, err
+	}
+
+	return t.Stat(path)
+}
+
+func GetFSForConnection(ctx *context.Context, c models.Connection) (connection.FilesystemRW, error) {
+	switch c.Type {
+	case models.ConnectionTypeAWS:
```
moshloop, line 194: I think we need a connection type file or folder

checks/runchecks.go
```diff
@@ -94,13 +99,74 @@ func RunChecks(ctx *context.Context) ([]*pkg.CheckResult, error) {
...
+	if err := saveArtifacts(ctx, results); err != nil {
+		logger.Errorf("error saving artifacts: %v", err)
+	}
+
 	return ProcessResults(ctx, results), nil
 }
 
+func saveArtifacts(ctx *context.Context, results pkg.Results) error {
+	if DefaultArtifactConnection == "" {
+		return nil
```
moshloop, line 114: Log a warning

checks/runchecks.go
```diff
@@ -94,13 +99,74 @@ func RunChecks(ctx *context.Context) ([]*pkg.CheckResult, error) {
...
+	}
+
+	fs, err := GetFSForConnection(utils.Ptr(ctx.Duty()), *connection)
+	if err != nil {
+		return fmt.Errorf("error getting filesystem for connection: %w", err)
+	}
+	defer fs.Close()
+
+	for _, r := range results {
+		if len(r.Artifacts) == 0 {
```
moshloop, line 131: We probably want to move this to the top of the function, of no artifacts have been returned don't even look for a connection

checks/runchecks.go
```diff
@@ -94,13 +99,74 @@ func RunChecks(ctx *context.Context) ([]*pkg.CheckResult, error) {
...
+			continue
+		}
+
+		checkIDRaw := ctx.Canary.Status.Checks[r.Check.GetName()]
```
moshloop, line 135:
```suggestion
		checkIDRaw := r.Check.GetPersistedID() or r.Canary.getPersistedId(check)
```

checks/runchecks.go
```diff
@@ -94,13 +99,74 @@ func RunChecks(ctx *context.Context) ([]*pkg.CheckResult, error) {
...
+				ConnectionID: connection.ID,
+				Path:         a.Path,
+				Filename:     info.Name(),
+				Size:         info.Size(),
+				ContentType:  a.ContentType,
+				Checksum:     hash.Sha256Hex(string(a.Content)),
+			}
+
+			if err := ctx.DB().Create(&artifact).Error; err != nil {
+				logger.Errorf("error saving artifact to db: %v", err)
```
moshloop, line 162: this should be a check failure

checks/connection_filesystem.go
```diff
@@ -0,0 +1,367 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	defer results.Body.Close()
+
+	return results.Body, nil
+}
+
+func (t *s3FS) Write(ctx gocontext.Context, path string, data []byte) (os.FileInfo, error) {
```
moshloop, line 108: I don't think we should use []byte as it will consume a lot of memory with large files - rather use io.Reader as the interface and an io.TeeReader with a sha checksum (which is also an io.Writer)  to calc the checksum

checks/exec.go
```diff
@@ -224,9 +227,54 @@ func setupConnection(ctx *context.Context, check v1.ExecCheck, cmd *exec.Cmd) er
...
+				}
+				defer file.Close()
+
+				if artifact.Content, err = io.ReadAll(file); err != nil {
+					logger.Errorf("error reading file. path=%s; %w", path, err)
+					continue
+				}
+
+				artifact.Path = path
+				artifact.ContentType = http.DetectContentType(artifact.Content)
```
moshloop, line 272: I think we can just use a https://github.com/nanmu42/limitio LimitedWriter and the https://github.com/gabriel-vasile/mimetype library for this so that we buffer at most 512KB to feed into mimetype detection, for larger files we wont necessarily detect the mime type.

checks/runchecks.go
```diff
@@ -94,13 +104,90 @@ func RunChecks(ctx *context.Context) ([]*pkg.CheckResult, error) {
...
 	}
 
+	if err := saveArtifacts(ctx, results); err != nil {
+		logger.Errorf("error saving artifacts: %v", err)
+	}
+
 	return ProcessResults(ctx, results), nil
 }
 
+func saveArtifacts(ctx *context.Context, results pkg.Results) error {
```
moshloop, line 117: Can we extract this function to an artifacts.go file thay we can just copy and paste for incident commander ?

Issue #1463, opened 2023-11-23
invalid uuid error 


yashmehrotra:
```
Rank 	Times reported 	Error
1 	34,881

ERROR: invalid input syntax for type uuid: ...

ERROR: invalid input syntax for type uuid: "/dev/petclinic-cloud-mysql-checks/mysql:Ping Visits DB/Verify that the database used by the Visits microservice is up and functioning:768806b5b55a6640dad4ed39fbab0635"

Context: unnamed portal parameter $1 = '...'
Statement: WITH filtered_check_status AS ( SELECT * FROM check_statuses WHERE check_id = $1 ) SELECT checks.id::text, canary_id::text, stats.passed, stats.failed, stats.p99, stats.p97, stats.p95, statii, type, checks.icon, checks.name, checks.description, canaries.namespace, canaries.name as canaryName, canaries.labels || checks.labels as labels, severity, owner, last_runtime, checks.created_at, checks.updated_at, checks.deleted_at, status, stats.max_time, stats.min_time, stats.total_checks FROM checks checks RIGHT JOIN ( SELECT check_id, percentile_disc(0.99) within group (order by filtered_check_status.duration) as p99, percentile_disc(0.97) within group (order by filtered_check_status.duration) as p97, percentile_disc(0.05) within group (order by filtered_check_status.duration) as p95, COUNT(*) FILTER (WHERE filtered_check_status.status = TRUE) as passed, COUNT(*) FILTER (WHERE filtered_check_status.status = FALSE) as failed, COUNT(*) total_checks, MIN(filtered_check_status.time) as min_time, MAX(filtered_check_status.time) as max_time FROM filtered_check_status GROUP BY check_id ) as stats ON stats.check_id = checks.id INNER JOIN canaries on checks.canary_id = canaries.id RIGHT JOIN ( SELECT check_id, json_agg(json_build_object('status',status,'duration',duration,'time',time )) as statii FROM ( SELECT check_id, status, time, duration, message, error, rank() OVER ( PARTITION BY check_id ORDER BY time DESC ) FROM filtered_check_status ) check_statuses WHERE rank <= $2 GROUP by check_id ) as statuses ON statuses.check_id = checks.id WHERE (stats.passed > 0 OR stats.failed > 0) AND checks.id = $1

Date: 2023-11-23 09:30:45
Database: mission_control
Application: 
User: postgres
Remote: 10.0.4.120

[...]

PR #1472, opened 2023-11-28
fix: invalid uuid

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/1463

### DIFF
--- a/pkg/db/canary.go
+++ b/pkg/db/canary.go
@@ -152,7 +152,7 @@ func GetTransformedCheckIDs(ctx context.Context, canaryID string, excludeTypes .
 
 func LatestCheckStatus(ctx context.Context, checkID string) (*models.CheckStatus, error) {
 	var status models.CheckStatus
-	if err := ctx.DB().Where("check_id = ?", checkID).Order("created_at DESC").Find(&status).Error; err != nil {
+	if err := ctx.DB().Select("time, created_at, status").Where("check_id = ?", checkID).Order("created_at DESC").Find(&status).Error; err != nil {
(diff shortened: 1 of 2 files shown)


pkg/db/canary.go
```diff
@@ -145,6 +145,19 @@ func GetTransformedCheckIDs(ctx context.Context, canaryID string, excludeTypes .
 	return ids, err
 }
 
+func LatestCheckStatus(ctx context.Context, checkID string) (*models.CheckStatus, error) {
+	var status models.CheckStatus
+	if err := ctx.DB().Where("check_id = ?", checkID).Order("created_at DESC").Find(&status).Error; err != nil {
```
moshloop, line 150: Can we only return the status and time fields?

pkg/jobs/canary/canary_jobs.go
```diff
@@ -208,18 +208,19 @@ func updateCanaryStatusAndEvent(canary v1.Canary, results []*pkg.CheckResult) {
...
+		latestCheckStatus, err := db.LatestCheckStatus(ctx, checkID)
+		if err != nil || latestCheckStatus == nil {
 			transitioned = true
-		} else if len(lastStatus) > 0 && (lastStatus[0].Statuses[0].Status != result.Pass) {
+		} else if latestCheckStatus.Status != result.Pass {
 			transitioned = true
 		}
 		if transitioned {
-			checkStatus[checkKey].LastTransitionedTime = &metav1.Time{Time: time.Now()}
+			checkStatus[checkID].LastTransitionedTime = &metav1.Time{Time: time.Now()}
```
moshloop, line 223: We should be using the db now

Issue #990, opened 2023-05-16
Canary scaling subresource


moshloop:
Allow canary objects to be scaled to 0, allowing them to be disabled as part of scale to zero of underlying pods

moshloop:
e.g. to be used with Keda like

```yaml
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: api
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Canary
    name: api
  triggers:
    - type: cron
      metadata:
        start: 0 * * * *
        end: 10 * * * *
        desiredReplicas: 0
```

or via kubectl scale --replicas=0 canary object

It probably makes the most sense to add and remove  `suspend` annotation on the canary in the scale handler, and then support the suspend annotation normally

PR #1477, opened 2023-11-30
feat: canary scale subresource

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/990


moshloop:
@adityathebe What actually stops the canary from running? I think we need to update the pre-run db call to check if its suspended 

It might also make more sense to a suspend db column that is mapped to the annotation

adityathebe:
> What actually stops the canary from running

https://github.com/flanksource/canary-checker/blob/4fa6e09530378fff0fa3066073dc9e72b7392315/pkg/jobs/canary/canary_jobs.go#L333-L335

Before the job is scheduled, there's this Ignored check which looks at the annotation for `suspend: true`

moshloop:
@adityathebe I guess not, can we also support a scale to 2 replicas, which triggers an immiedate run of the check, and then reverts back to 1 ?

api/v1/canary_types.go
```diff
@@ -281,6 +285,10 @@ type CanaryStatus struct {
 	Latency1H string `json:"latency1h,omitempty"`
 	// used for keeping history of the checks
 	runnerName string `json:"-"`
+	// Replicas keep track of the number of replicas
+	Replicas int `json:"replicas,omitempty"`
+	// Selector ??
+	Selector map[string]string `json:"selector,omitempty"`
```
moshloop, line 291: I believe this is just a string

PR #1488, opened 2023-12-08
feat: allow forming relationships to component and configs by lookup

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/1444

wait for: https://github.com/flanksource/duty/pull/358


pkg/jobs/canary/canary_jobs.go
```diff
@@ -170,6 +181,99 @@ func (j CanaryJob) Run(ctx dutyjob.JobRuntime) error {
...
+		return fmt.Errorf("error parsing canary id(%s): %w", result.Canary.GetCheckID(result.Check.GetName()), err)
+	}
+
+	canaryID, err := uuid.Parse(result.Canary.GetPersistedID())
+	if err != nil {
+		return fmt.Errorf("error parsing canary id(%s): %w", result.Canary.GetPersistedID(), err)
+	}
+
+	for _, lookup := range relationshipConfig.Components {
+		if populated, err := lookup.Populate(result.Labels, map[string]any{"result": result}); err != nil {
```
adityathebe, line 217: `map[string]any{"result": result}`

@moshloop is this okay for the cel's env var?

pkg/jobs/canary/canary_jobs.go
```diff
@@ -170,6 +181,99 @@ func (j CanaryJob) Run(ctx dutyjob.JobRuntime) error {
...
+
+		logger.Infof("Finding all components with %s/%s/%s", lookup.Namespace.Value, lookup.Name.Value, lookup.Type.Value)
+		componentIDs, err := db.FindComponentIDsByNameNamespaceType(ctx, lookup.Namespace.Value, lookup.Name.Value, lookup.Type.Value)
+		if err != nil {
+			logger.Errorf("error finding components (check=%s) (lookup=%v): %v", checkID, lookup, err)
+			continue
+		}
+
+		for _, componentID := range componentIDs {
+			selectorID, err := utils.GenerateJSONMD5Hash(check)
```
adityathebe, line 232: Not sure if this is appropriate to generate the relationship selector id

moshloop: Should the selector Id be a hash of the selector spec ?

api/v1/checks.go
```diff
@@ -481,10 +483,113 @@ type AlertManager struct {
 	AlertManagerCheck `yaml:",inline" json:",inline"`
 }
 
+// CheckRelationshipLookup specifies the type of lookup to perform.
+type CheckRelationshipLookup struct {
```
moshloop, line 487: Can we make this + the Name, Namespace, Type tuple a generic lookup struct in duty, and then we can have
```go
type TemplateEnvironment interface {
GetEnvironment() map[string]any
}

duty.LookupComponent(ctx,  lookup)
duty.LookupConfig(ctx,  lookup)
```

We can then eventually replace all the existing lookup type use cases with the common Lookup struct and then have the option to add more type of matchers in the future that would apply everywhere?

adityathebe: https://github.com/flanksource/duty/pull/359

PR #1491, opened 2023-12-12
chore: add otel_collector to chart

yashmehrotra:


chart/templates/deployment.yaml
```diff
@@ -161,6 +161,9 @@ spec:
             {{- if ne .Values.prometheusURL "" }}
             - --prometheus={{ .Values.prometheusURL }}
             {{- end }}
+            {{- if .Values.otel_collector }}
+            - --otel-collector-url={{ tpl .Values.otel_collector . | default "" }}
```
yashmehrotra, line 165: Using `tpl` so `{{ .Values.key }}` can be used to resolve it

PR #1495, opened 2023-12-14
fix: improve nil handling of sql/folder checks

moshloop:


adityathebe:
All changes are in master

PR #1496, opened 2023-12-15
chore: cache FindConfig function

yashmehrotra:


pkg/db/config_db.go
```diff
@@ -47,20 +50,40 @@ func configQuery(config pkg.Config) *gorm.DB {
 	return query
 }
 
+var configCache = gocache.New(30*time.Minute, 1*time.Hour)
```
yashmehrotra, line 53: Is cache time of 30 minutes too high ? or works ?

PR #1533, opened 2023-12-27
feat: move check details query to duty

adityathebe:


adityathebe:
@moshloop This is ready

PR #1556, opened 2024-01-03
fix: transformed checks returning correct id

yashmehrotra:


pkg/cache/postgres.go
```diff
@@ -58,7 +58,7 @@ func (c *postgresCache) AddCheckFromStatus(check pkg.Check, status pkg.CheckStat
 	}
 
 	if check.ID != uuid.Nil {
-		return uuid.Nil, nil
```
yashmehrotra, line 61: @moshloop Not sure why this was set to nil uuid

Due to this being nil id, existing checks were marked deleted instead of being updated

PR #1587, opened 2024-01-15
fix: deprecate templatebody from http check

### LINKED ISSUE #1584: Deprecate the `templateBody` field in HTTP check spec
we always template it.
Users can perform escaping https://stackoverflow.com/a/17642427/6199444

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/1584

### DIFF
--- a/api/v1/checks.go
+++ b/api/v1/checks.go
@@ -196,7 +196,8 @@ type HTTPCheck struct {
 	Body string `yaml:"body,omitempty" json:"body,omitempty" template:"true"`
 	// Header fields to be used in the query
 	Headers []types.EnvVar `yaml:"headers,omitempty" json:"headers,omitempty"`
-	// Template the request body
+	// Template the request body.
+	// Deprecated: Body is always templated by default. Use template escaping if necessary.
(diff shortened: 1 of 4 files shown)


adityathebe:
To avoid excessive escaping on the templates we might want to simply turn off templating on some fields. Example: we could do it via annotations


```yaml
kind: Canary
metadata:
  name: http-fail
  labels:
    "Expected-Fail": "true"
  annotations:
    canary.template.disabled: 'http.body, http.url' # <-----
spec:
  interval: 30
  http:
    - endpoint: https://httpbin.demo.aws.flanksource.com/status/500
      name: http fail response code check
      responseCodes: [200]
    - endpoint: https://httpbin.demo.aws.flanksource.com/status/200
      name: http fail test expr check
      display:
        expr: string(code) + " should be 500"
      test:
        expr: code == 500
```

Should this be sufficient

```
canary.template.disabled: 'http.body, http.url' # <-----
```

or do we want to individually disable templating on the http check. Eg

```yaml

adityathebe:
It's been a while since this PR. Not sure if we need it.

PR #1623, opened 2024-01-29
fix: windows service script & postgres shutdown

junaid-ebrahim:
2 Main Fixes

1. Fixed windows powershell script by doing the following
  1.1 Removed deprecated options
  1.2 Added Postgres local storage location creation
  1.3 Added skip DB migrations option to prevent crash
  1.4 Added parameter to install, uninstall & reinstall to make these operations easier
  1.5 Added ability to use local canary-checker executable if one exists, prevent needless download & allow for installing custom build

2. Fixed Postgres not shutdown, this is due to log.fatal after Postgres started. Fix is to simply explicitly call shutdown before these statements.

### DIFF
--- a/install-service.ps1
+++ b/install-service.ps1
@@ -1,21 +1,32 @@
  param (
     [string]$configfile = "$pwd\canary-checker.yaml",
     [int]$httpPort = 8080,
-    [int]$metricsPort = 8081,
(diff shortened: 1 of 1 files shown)


junaid-ebrahim:
Rebased with master

PR #1646, opened 2024-02-02
chore: optimize config and component relationship queries

yashmehrotra:


pkg/topology/component_config.go
```diff
@@ -44,12 +43,19 @@ var ComponentConfigRun = &job.Job{
 		}
 
 		// Cleanup dead relationships
-		componentIDsWithConfigs := lo.Map(components, func(c pkg.Component, _ int) string { return c.ID.String() })
-		if err := db.Table("config_component_relationships").
-			Where("component_id NOT IN ?", componentIDsWithConfigs).
-			Update("deleted_at", duty.Now()).Error; err != nil {
+		// TODO: Agent_id matters ? Since logic is same
```
yashmehrotra, line 46: @moshloop Should we keep agent_id in the query ? One train of thought is that an instance should only process its own resources, the other option is that since the logic is same across instances it shouldn't matter

What do you think ?

yashmehrotra: Removed agent_id filter

PR #1649, opened 2024-02-02
fix: topology multi scheduling bug

yashmehrotra:


pkg/jobs/topology/topology_jobs.go
```diff
@@ -81,7 +81,7 @@ var SyncTopology = &job.Job{
 func SyncTopologyJob(ctx context.Context, t pkg.Topology) error {
 	id := t.ID
 	var existingJob *job.Job
-	if j, ok := topologyJobs.Load(id); ok {
+	if j, ok := topologyJobs.Load(id.String()); ok {
```
yashmehrotra, line 84: We were loading uuid but storing strings

pkg/jobs/topology/topology_jobs.go
```diff
@@ -102,7 +102,7 @@ func SyncTopologyJob(ctx context.Context, t pkg.Topology) error {
 	}
 
 	existingTopology := existingJob.Context.Value("topology")
-	if existingTopology != nil && !reflect.DeepEqual(existingTopology.(v1.Topology).Spec, t.Spec) {
+	if existingTopology != nil && !reflect.DeepEqual(existingTopology.(v1.Topology).Spec, v1Topology.Spec) {
```
yashmehrotra, line 105: comparing v1 spec with v1 spec, previously it was v1 spec and db's duty.json type

PR #1660, opened 2024-02-07
chore: standardize agent chart values

yashmehrotra:


chart/templates/deployment.yaml
```diff
@@ -141,7 +141,6 @@ spec:
             - --{{$k}}={{$v}}
             {{- end }}
             {{- if .Values.upstream.enabled }}
-            - --agent-name={{ .Values.upstream.agentName }}
```
moshloop, line 144: Why do we remove this?

yashmehrotra: Made a change to accomodate both

PR #1668, opened 2024-02-13
feat: run check endpoint

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/1665

### DIFF
--- a/pkg/api/run_now.go
+++ b/pkg/api/run_now.go
@@ -4,13 +4,18 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/flanksource/canary-checker/api/context"
+	v1 "github.com/flanksource/canary-checker/api/v1"
 	"github.com/flanksource/canary-checker/checks"
 	"github.com/flanksource/canary-checker/pkg"
+	"github.com/flanksource/canary-checker/pkg/cache"
 	"github.com/flanksource/canary-checker/pkg/db"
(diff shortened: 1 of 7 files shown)


pkg/api/run_now.go
```diff
@@ -20,59 +25,52 @@ type CheckErrorMessage struct {
...
-	if err != nil {
+	for _, result := range results {
+		_ = cache.PostgresCache.Add(pkg.FromV1(result.Canary, result.Check), pkg.CheckStatusFromResult(*result))
+		if err := canaryJobs.FormCheckRelationships(ctx.Context, result); err != nil {
+			ctx.Logger.Named(result.Name).Errorf("error forming check relationships: %v", err)
+		}
+	}
+
+	canaryJobs.UpdateCanaryStatusAndEvent(ctx.Context, *canary, results)
+	canaryJobs.CanaryLastRuntimes.Store(canary.GetUID(), time.Now())
```
moshloop, line 67: I don't think we should impact with normal scheduling 
```suggestion
```

pkg/api/run_now.go
```diff
@@ -20,59 +25,52 @@ type CheckErrorMessage struct {
...
+
+	var response models.CheckSummary
+	if err := ctx.DB().Where("id = ?", checkID).First(&response).Error; err != nil {
```
moshloop, line 70: I don't think this will be deterministic - another check could have been inserted I think we can just return the results object

PR #1687, opened 2024-02-26
chore: use context templating

yashmehrotra:


pkg/topology/context.go
```diff
@@ -38,23 +38,17 @@ func (c *ComponentContext) GetTemplater() gomplate.StructTemplater {
...
-		},
-		Values: map[string]interface{}{
-			"component":  c.CurrentComponent,
-			"properties": props,
-		},
+	vals := map[string]any{
+		"component":  c.CurrentComponent,
+		"properties": props,
 	}
+	// TODO: This has delims as ${ }
```
yashmehrotra, line 49: @moshloop This is the only place in codebase that uses `${ }` delims

Should I keep it as is or use the default ones here

pkg/topology/context.go
```diff
@@ -38,23 +38,17 @@ func (c *ComponentContext) GetTemplater() gomplate.StructTemplater {
...
 		return *c.templater
 	}
-	var props = make(map[string]interface{})
+	var props = make(map[string]any)
 	if c.CurrentComponent != nil && c.CurrentComponent.Properties != nil {
 		props = c.CurrentComponent.Properties.AsMap()
 	}
-	c.templater = &gomplate.StructTemplater{
-		// RequiredTag: "template",
-		DelimSets: []gomplate.Delims{
```
yashmehrotra, line 47: @yashmehrotra Test failing is in the fixture `component-with-for-each.yml`

PR #1707, opened 2024-03-11
fix: component lookups

adityathebe:


pkg/topology/run.go
```diff
@@ -31,30 +32,37 @@ func mergeComponentLookup(ctx *ComponentContext, component *v1.ComponentSpec, sp
...
-		// the property returned a list of new properties
-		for _, result := range results {
-			var p pkg.Component
-			data, err := json.Marshal(result)
-			if err != nil {
-				return nil, fmt.Errorf("error marshaling result to json: %w", err)
-			}
-			if err := json.Unmarshal(data, &p); err != nil {
-				return nil, fmt.Errorf("error unmarshaling data from json: %w", err)
+		switch v := results[0].(type) {
```
moshloop, line 37: Can't this cause an index out of bounds?

adityathebe: There's a length check above it

PR #1711, opened 2024-03-13
fix: sql check scanner

adityathebe:
The sql checks attempts to scan columns of all types as a string.

This PR adds better supports for scanning columns to appropriate native Go types.
This is helpful in the templates as you don't have to cast anything.


checks/sql.go
```diff
@@ -20,7 +23,7 @@ type SQLChecker interface {
 
 type SQLDetails struct {
 	Rows  []map[string]interface{} `json:"rows"`
-	Count int                      `json:"count,omitempty"`
+	Count int                      `json:"count"`
```
adityathebe, line 26: don't omit it otherwise the templates need to perform nil check first.

Issue #436, opened 2022-03-24
Kubernetes Generic Check


moshloop:
```yaml
kubernetesResource:
        
    - name: deploy a pod and check that it is accessible via ingress
       staticResources: # are deleted when the canary is deleted
         - Ingress spec
         - svc spec
       resources: # deleted after the check
         - A pod spec
       waitForReady: true # wait for resources to be ready before running checks
       timeout: 10m
       testResults: # extract test results from running pods
            -  path: /test/junit.xml
                type: junit # default
                pod:
                    name: {{.resources[0].metadata.name
                    # selector:
       checks:
         - http:
             url: {{.staticResources[0].ingress.spec.hostname }}             
```

Successor to pod, namespace jmeter and junit checks 

Work started here: https://github.com/flanksource/canary-checker/pull/368

k8sexperiment:
It would be very cool to have a full, flexible Kubernetes check specification, especially for cloud environments:

Create a pod, and once it's ready, you can check:

External-dns/Ingress/Cert-manager via Ingress manifest: Once the Ingress is created, you can verify that a certificate was created for the host and a record was created via external-dns. Hence, if TLS ingress is successful, the URL check should return a 200/green status.

ImagePull: This is simple; if the pod is created and pulled using a custom registry, it should be working properly.

HPA/Keda: Utilizing k6 load test, check if the pod is scaling based on metrics.

WorkloadIdentity: Suppose correct credentials are added as annotations/labels to the pod's ServiceAccount and pod.spec.template. In that case, you might want to test access to the Cloud resources from the pod. You can mount secrets from a cloud vault or push/pull files from S3/GCS/Azure StorageAccount, etc. Again, this might be related to Kubernetes resources specification.

PVC Volume attached: Test CSI to ensure proper functionality.

PR #1756, opened 2024-03-28
feat: new kubernetes resource check

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/436


checks/kubernetes_resource.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	for i := range check.Resources {
+		resource := check.Resources[i]
+		resource.SetAnnotations(map[string]string{annotationkey: ctx.Canary.ID()})
+		if err := ctx.Kommons().ApplyUnstructured(ctx.Namespace, &resource); err != nil {
+			return results.Failf("failed to apply resource %s: %v", resource.GetName(), err)
+		}
+
+		defer func() {
```
moshloop, line 116: I don't think this should be defered - A successful delete is part of the test

adityathebe: Even if it's deferred, we mark the check as failed in case of error during deletion.

fixtures/k8s/kubernetes_resource.yaml
```diff
@@ -0,0 +1,28 @@
...
+      waitForReady: true 
+      timeout: 10m
+      resources:
+        - apiVersion: v1
+          kind: Pod
+          metadata:
+            name: nginx-pod
+            namespace: default
+          spec:
+            containers:
```
moshloop, line 24: Need to add a static ingress and svc + the http check (using the internal nginx service and a host header to match so the dns doesn't need to be setup)

fixtures/k8s/kubernetes_resource_service.yaml
```diff
@@ -0,0 +1,45 @@
+---
+apiVersion: canaries.flanksource.com/v1
+kind: Canary
+metadata:
+  name: pod-svc-test
+  namespace: default
+  labels:
+    "Expected-Fail": "false"
```
moshloop, line 8: Can we hadd pass/fail variants with the suffixes ?

checks/kubernetes_resource.go
```diff
@@ -0,0 +1,193 @@
...
+	"github.com/flanksource/duty/types"
+
+	"github.com/flanksource/canary-checker/api/context"
+	v1 "github.com/flanksource/canary-checker/api/v1"
+	"github.com/flanksource/canary-checker/pkg"
+)
+
+const (
+	// maximum number of static & non static resources a canary can have
+	maxResourcesAllowed = 10
```
moshloop, line 24: Can we make this a system property ?

api/v1/checks.go
```diff
@@ -776,6 +780,109 @@ func (c ConfigDBCheck) GetEndpoint() string {
...
+
+	// Timeout to wait for all static & non-static resources to be ready.
+	// 	Default: 10m
+	Timeout string `json:"timeout,omitempty"`
+
+	// Interval to check if all static & non-static resources are ready.
+	// 	Default: 30s
+	Interval string `json:"interval,omitempty"`
+
+	parsedTimeout  *time.Duration
```
moshloop, line 808: I think these need empty json tags

checks/kubernetes_resource.go
```diff
@@ -0,0 +1,223 @@
...
+
+	var attempts int
+	backoff := retry.WithMaxDuration(waitTimeout, retry.NewConstant(waitInterval))
+	retryErr := retry.Do(ctx, backoff, func(_ctx gocontext.Context) error {
+		ctx = _ctx.(*context.Context)
+		attempts++
+		ctx.Tracef("waiting for %d resources to be ready. (attempts: %d)", check.TotalResources(), attempts)
+
+		var templateVar = map[string]any{}
+		kClient := pkg.NewKubeClient(ctx.Kommons().GetRESTConfig)
```
moshloop, line 197: This is an expensive operation - why can we  just use the ctx.Kubernetes() ?

checks/kubernetes_resource.go
```diff
@@ -0,0 +1,275 @@
...
+		if check.CheckRetries.MaxRetries > 0 {
+			backoff = retry.WithMaxRetries(uint64(check.CheckRetries.MaxRetries), backoff)
+		}
+
+		if maxRetryTimeout, _ := check.CheckRetries.GetTimeout(); maxRetryTimeout > 0 {
+			backoff = retry.WithMaxDuration(maxRetryTimeout, backoff)
+		}
+
+		retryErr := retry.Do(ctx, backoff, func(_ctx gocontext.Context) error {
+			ctx.Infof("running check: %s", virtualCanary.Name)
```
moshloop, line 140: `ctx.Logger.V(4)`

checks/kubernetes_resource.go
```diff
@@ -0,0 +1,275 @@
...
+
+	var attempts int
+	backoff := retry.WithMaxDuration(waitTimeout, retry.NewConstant(waitInterval))
+	if check.WaitFor.MaxRetries > 0 {
+		backoff = retry.WithMaxRetries(uint64(check.WaitFor.MaxRetries), backoff)
+	}
+	retryErr := retry.Do(ctx, backoff, func(_ctx gocontext.Context) error {
+		ctx = _ctx.(*context.Context)
+		attempts++
+		ctx.Tracef("waiting for %d resources to be ready. (attempts: %d)", check.TotalResources(), attempts)
```
moshloop, line 186: `ctx.Logger.V(5)`

fixtures/k8s/kubernetes_resource_ingress_pass.yaml
```diff
@@ -0,0 +1,75 @@
+---
```
moshloop, line 1: Can we add an example that creates a pod, and waits for it to terminate and then using test check that the terminate exit code is 0

fixtures/k8s/kubernetes_resource_pod_exit_code_pass.yaml
```diff
@@ -0,0 +1,44 @@
...
+          metadata:
+            name: hello-world
+            namespace: default
+          spec:
+            restartPolicy: Never
+            containers:
+              - name: hello-world
+                image: hello-world
+      waitFor:
+        expr: 'dyn(resources).all(r, k8s.getHealth(r).status == "Healthy")'
```
adityathebe, line 27: should be replaced by `k8s.isReady` once we add it

fixtures/k8s/kubernetes_resource_pod_exit_code_pass.yaml
```diff
@@ -0,0 +1,44 @@
...
+            - name: exit-code-check
+              kind: Pod
+              namespaceSelector:
+                name: default
+              resource:
+                name: hello-world
+              test:
+                expr: >
+                  size(results) == 1 &&
+                  results[0].Object.status.containerStatuses[0].state.terminated.exitCode == 0
```
adityathebe, line 44: could probably add a helper function in cel to get containerStatus

checks/runchecks.go
```diff
@@ -262,15 +263,14 @@ func processTemplates(ctx *context.Context, r *pkg.CheckResult) *pkg.CheckResult
...
 		if err != nil {
 			r.ErrorMessage(err)
-		} else if message != "true" {
-			if message != "false" {
-				r.Failf("expecting either 'true' or 'false' but got '%v'", message)
-			} else {
-				r.Failf("")
-			}
+		} else if parsed, err := strconv.ParseBool(message); err != nil {
+			r.Failf("text expression did not return a boolean value. got %s", message)
```
moshloop, line 271:
```suggestion
			r.Failf("test expression did not return a boolean value. got %s", message)
```

checks/runchecks.go
```diff
@@ -262,15 +263,14 @@ func processTemplates(ctx *context.Context, r *pkg.CheckResult) *pkg.CheckResult
...
+		} else if !parsed {
+			r.Failf("test failed")
```
moshloop, line 273: Why are we adding this in ?

adityathebe: without this, the check was passing for me. It's probably because we have `if Error != ""` check somewhere.

moshloop: I think we should identify that place - This will add unnecessary messages into the tables  / UI

moshloop (approved):

PR #1767, opened 2024-04-02
fix: Update deployment.yaml for embedded DB option

connected-me:
_Appears to by resolved as of 1.0.258, though an upgrade in place does not seem to resolve._

**Helm chart release version(s):** 1.0.257 - 1.0.258-beta.9
**Issue:**
Enbedded postgres server created with:
postgres using /opt/database/bin/bin/pg_ctl start -w -D /opt/database/data -o -p 6432:

URL previously pointed to /opt/database/ causing crash reboot

**Steps to recreate:**
use the following values.yaml with the above versions
```
db:
  runMigrations: true
  embedded:
  # If the database is embedded, setting this to true will persist the contents of the database
  # through a persistent volume
    persist: true
    storageClass: ""
    storage: 20Gi
```

**Suggestions:**
update default embedded database URL to /opt/database/data

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -93,7 +93,7 @@ spec:
                   name: {{ if .Values.global.db.connectionPooler.enabled }}{{ .Values.global.db.connectionPooler.secretKeyRef.name }}{{ else }}{{ .Values.db.external.secretKeyRef.name }}{{ end }}
                   key: {{ if .Values.global.db.connectionPooler.enabled }}{{ .Values.global.db.connectionPooler.secretKeyRef.key }}{{ else }}{{ .Values.db.external.secretKeyRef.key }}{{ end }}
               {{- else }}
-              value: "embedded:///opt/database/"
+              value: "embedded:///opt/database/data/"
               {{- end }}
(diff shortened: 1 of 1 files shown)


moshloop:
Hi @connected-me - Thanks for the contribution - We are in the process of setting up some helm tests (which should expose this issue) after which we can merge

connected-me:
> @connected-me When I set it up using the helm chart
> 
> ```
> helm install canary-checker flanksource/canary-checker -n "canary-checker"  --create-namespace
> ```
> 
> it worked out of the box for me
> 
> Can you share your exact setup steps so that we can reproduce it

Will update the initial request but when including the values.yaml with the following settings

db:
  runMigrations: true
  embedded:
    # If the database is embedded, setting this to true will persist the contents of the database
    # through a persistent volume
    persist: true
    storageClass: "<your storage class>"
    storage: 20Gi

It's also possible that it doesn't chmod the folder to 0700 or 0750 but didn't want to dive too far down the rabbit hole.

yashmehrotra:
>edit: looks like it has been fixed in the most recent version already, can be disregarded.

> Will update the initial request but when including the values.yaml in the previous version (updated in initial comment) with the following settings

Cool, closing this PR

Issue #1782, opened 2024-04-11
New ingress test check fails 1 out of 2


kubebn:
https://github.com/flanksource/canary-checker/blob/master/fixtures/k8s/kubernetes_resource_ingress_pass.yaml

![image](https://github.com/flanksource/canary-checker/assets/110397870/0c5a2d32-12de-4ce7-9f58-8703c82647a0)

Error: `failed to apply resource httpbin: Pod/canary-checker/httpbin: pods "httpbin" already exists`

PR #1790, opened 2024-04-15
fix: better deletion of resources

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/1782

### DIFF
--- a/checks/kubernetes_resource.go
+++ b/checks/kubernetes_resource.go
@@ -2,21 +2,25 @@ package checks
 
 import (
 	gocontext "context"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/flanksource/gomplate/v3"
 	"github.com/sethvargo/go-retry"
-
-	"github.com/flanksource/commons/logger"
-	"github.com/flanksource/commons/utils"
-	"github.com/flanksource/duty/types"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/watch"
 
 	"github.com/flanksource/canary-checker/api/context"
 	v1 "github.com/flanksource/canary-checker/api/v1"
 	"github.com/flanksource/canary-checker/pkg"
+	"github.com/flanksource/commons/utils"
+	"github.com/flanksource/duty/types"
 )
 
 const (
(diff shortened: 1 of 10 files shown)


checks/kubernetes_resource.go
```diff
@@ -277,3 +299,58 @@ func (c *KubernetesResourceChecker) validate(ctx *context.Context, check v1.Kube
...
+// deleteResourceSync deletes the given kubernetes resource and waits for it to be deleted.
+func deleteResourceSync(ctx *context.Context, resource unstructured.Unstructured) error {
+	ctx.Logger.V(5).Infof("deleting resource (%s/%s/%s)", resource.GetKind(), resource.GetNamespace(), resource.GetName())
+
+	rc, err := ctx.Kommons().GetRestClient(resource)
+	if err != nil {
+		return fmt.Errorf("failed to get rest client config for (%s/%s/%s): %w", resource.GetKind(), resource.GetNamespace(), resource.GetName(), err)
+	}
+
+	watcher, err := rc.Watch(resource.GetNamespace(), resource.GetAPIVersion(), &metav1.ListOptions{})
```
moshloop, line 312: I don't think watches are meant to be used like this - they are fairly expensive on etcd. 

Simple polling should be better (but we should probably have a waitForDelete option, if the resource names can be made unique with a random value, then we don't need to wait)

For deletes the RestClient can be used - Make sure to specify GracePeriodSeconds: 0 and propagationPolicy: orphan on the DeleteOptions

checks/kubernetes_resource.go
```diff
@@ -277,3 +299,54 @@ func (c *KubernetesResourceChecker) validate(ctx *context.Context, check v1.Kube
...
+
+		return fmt.Errorf("failed to delete resource %s: %w", resource.GetName(), err)
+	}
+
+	if !waitForDelete {
+		return nil
+	}
+
+	// Poll the resource until it's deleted
+	ticker := time.NewTicker(2 * time.Second)
```
moshloop, line 333: This should probably be split up, delete all the resources, then poll all the resources otherwise we wait for 1 resource to be deleted, before deleting the next,

moshloop: Which means we will poll all resources concurrently -  If we delete all resources first and then start polling:

1) If there are many resources, the first resources may have already been deleted by the time we start polling
2) After finding the first non-deleted resources we can sleep instead of polling them all, as the resources are deleted, remove them from the list of polling and only check the first item in the list-

checks/kubernetes_resource.go
```diff
@@ -277,3 +293,86 @@ func (c *KubernetesResourceChecker) validate(ctx *context.Context, check v1.Kube
...
+	ctx.Logger.V(4).Infof("deleting %d resources", len(resources))
+
+	// cache dynamic clients
+	clients := map[string]*cliresource.Helper{}
+
+	eg, _ := errgroup.WithContext(ctx)
+	for i := range resources {
+		resource := resources[i]
+		eg.Go(func() error {
+			rc, err := ctx.Kommons().GetRestClient(resource)
```
moshloop, line 307: This should be cache based on apiVersion and kind

moshloop:
@adityathebe  I don't think we need a waitFor.maxRetries as we have a timeout, and the docs for the default on interval say 30s instead of 5s

Issue #1764, opened 2024-04-02
Remove configDb / ec2  checks


adityathebe:
```sql
SELECT * FROM config_items WHERE config_class = 'CertificateRequest' AND config->'metadata'->'annotations' @> '{"cert-manager.io/certificate-name": "homelab"}'
```

This query is parsed parsed to validate that the query only calls `config_` tables. However that parser returns an in valid query error.

I tried bypassing that validation and the query works.

PR #1798, opened 2024-04-19
delete deprecated checks

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/1764


yashmehrotra:
@adityathebe @moshloop Can we merge this ? Since latest duty does not have `config/query` funcs, it cant be updated in canary checker

PR #1803, opened 2024-04-20
feat: template resources in kubernetes resource check

adityathebe:
we allow templating everything including the name of both static and non-static resources except
- namespace
- Group 
- Version
- Kind

---

All polling and deleting operations, which were name based, have been moved to use the namespace + GroupVersionKind api.

If the program crashes right after it has created the kubernetes resources but before it has deleted them, then it'll just clean those resources up in the next run even if the name of the resources are randomly generated.

---

These labels are added to all the resources

```
canaries.flanksource.com/canary-id=d789b028-d3e6-43f7-9898-f9c5a3bf2f4e
canaries.flanksource.com/check-id=018f0ae6-030a-f9bd-5e78-c6e060394957
```

additionally, static resources also gets

```
canaries.flanksource.com/is-static=true
```

### DIFF
--- a/checks/kubernetes_resource.go
+++ b/checks/kubernetes_resource.go
@@ -10,6 +10,8 @@ import (
 	"time"
 
 	"github.com/flanksource/gomplate/v3"
+	"github.com/flanksource/is-healthy/pkg/health"
+	"github.com/flanksource/is-healthy/pkg/lua"
 	"github.com/samber/lo"
 	"github.com/sethvargo/go-retry"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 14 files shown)


checks/kubernetes_resource.go
```diff
@@ -95,7 +99,6 @@ func (c *KubernetesResourceChecker) Check(ctx *context.Context, check v1.Kuberne
 
 	for i := range check.Resources {
 		resource := check.Resources[i]
-
 		resource.SetAnnotations(map[string]string{annotationkey: ctx.Canary.ID()})
```
moshloop, line 102: We need to use check id vs canary-id as there could be more than 1 check per canary

I think a label might be better - we can then just track the kinds, do a single `kubectl delete {kind} --all-namespaces -l ` - If you remove a kind a kind from an existing check then it gets orphaned.

fixtures/k8s/kubernetes_resource_ingress_pass.yaml
```diff
@@ -13,6 +13,7 @@ spec:
       namespace: default
       description: "deploy httpbin & check that it's accessible via ingress"
       waitFor: 
+        expr: 'dyn(resources).all(r, k8s.isReady(r))'
```
moshloop, line 16: do we need the `dyn` ?

adityathebe: yes, it's necessary.


```
ERROR: <input>:1:1: expression of type 'any' cannot be range of a comprehension (must be list, map, or dynamic)
 | resources.all(r, k8s.isReady(r))
 | ^
```

checks/kubernetes_resource.go
```diff
@@ -282,45 +288,92 @@ func (c *KubernetesResourceChecker) validate(ctx *context.Context, check v1.Kube
...
-				if errors.As(err, &statusErr) {
-					switch statusErr.ErrStatus.Code {
-					case 404:
-						return nil
+
+			switch resource.GetKind() {
+			case "Namespace", "Service":
+				// NOTE: namespace cannot be deleted with `.DeleteCollection()`
+				//
+				// Even though Service can be deleted with `.DeleteCollection()`
```
moshloop, line 340:
```suggestion
				// FIXME: Even though Service can be deleted with `.DeleteCollection()`
```

checks/kubernetes_resource.go
```diff
@@ -282,45 +288,92 @@ func (c *KubernetesResourceChecker) validate(ctx *context.Context, check v1.Kube
...
-			gvk := resource.GetObjectKind().GroupVersionKind()
-			clients.Store(gvk, rc)
+			cachedClient, _ := clients.Load(resource.GetObjectKind().GroupVersionKind())
+			rc := cachedClient.(dynamic.ResourceInterface)
 
-			namespace := utils.Coalesce(resource.GetNamespace(), ctx.Namespace)
-			deleteOpt := &metav1.DeleteOptions{
+			deleteOpt := metav1.DeleteOptions{
 				GracePeriodSeconds: lo.ToPtr(int64(0)),
 				PropagationPolicy:  lo.ToPtr(metav1.DeletePropagationOrphan),
```
moshloop, line 333: If waitForDelete, then it should be DeletePropogationForeground, otherwise DeletePropogationBackground

PR #1851, opened 2024-05-20
chore: update chart log values to allow templating

yashmehrotra:


chart/values.yaml
```diff
@@ -36,7 +36,8 @@ debug: false
 
 # -v, -vv, -vvv
 logLevel: ""
-jsonLogs: true
+# This has to be stringy because of the way templates get evaluated
+jsonLogs: "true"
```
moshloop, line 40: I am not sure we want to expose this "hack"  - why not just use global.jsonLogs ?

yashmehrotra: Using `tpl` is the only way to template values from the same file

If in the mission-control chart, we want to `global.jsonLogs` to be passed here, we have to use `tpl` but that does not work with bools

yashmehrotra: We can have something like this: https://github.com/Masterminds/sprig/issues/258

but thats even more hacky

PR #1881, opened 2024-05-27
feat: support grouping topologies by tags

adityathebe:
resolves: https://github.com/flanksource/mission-control-registry/issues/162


moshloop:
@adityathebe can we add a test for this ?

adityathebe:
@moshloop it's caused by the missing `uid` on the fixtures. Can be ignored.

PR #1886, opened 2024-05-28
feat: save check & status in a single transaction

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1033


pkg/cache/postgres.go
```diff
@@ -29,30 +31,40 @@ func NewPostgresCache(context context.Context) *postgresCache {
 	}
 }
 
+func (c *postgresCache) saveCheckAndStatus(check pkg.Check, status pkg.CheckStatus) (string, error) {
+	tx := c.DB().Begin()
```
moshloop, line 35: Why not do this for the entire list of checks ? a tx per check status is probably overkill

PR #1898, opened 2024-06-12
fix: prometheus auth in check

yashmehrotra:


pkg/prometheus/prometheus.go
```diff
@@ -43,37 +42,37 @@ func (p PrometheusClient) GetUptime(checkKey, duration string) (float64, error)
...
+		if !auth.Username.IsEmpty() {
 			roundTripper = prometheusConfig.NewBasicAuthRoundTripper(
 				username,
 				prometheusConfig.Secret(password),
 				"",
 				"",
 				roundTripper)
-		} else if auth.Oauth != nil {
-			clientID, err := ctx.GetEnvValueFromCache(auth.Oauth.ClientID, ctx.GetNamespace())
+		} else if !auth.OAuth.IsEmpty() {
```
moshloop, line 69: I think this if is too nested, it requires username to be not empty (which triggers the higher if) and this condition is never called.

yashmehrotra: Simplified to single nesting

PR #1906, opened 2024-06-16
chore: use ecr registry and hardcode service account

yashmehrotra:


chart/templates/postgres-statefulset.yaml
```diff
@@ -20,7 +20,7 @@ spec:
     spec:
       containers:
       - name: postgresql
-        image: postgres:14
+        image: public.ecr.aws/k4y9r6y5/postgres:14.1.0.21
```
moshloop, line 23: Shouldn't this be global registry ?

chart/values.yaml
```diff
@@ -3,6 +3,7 @@
 # Declare variables to be passed into your templates.
 
 global:
+  imageRegistry: public.ecr.aws/k4y9r6y5
```
moshloop, line 6: lets keep the default in the charts to docker and set it to ECR only in the mission-control chart

.github/workflows/helm-test.yml
```diff
@@ -47,6 +47,17 @@ jobs:
       - name: Install helm chart
         run: 'helm install canary-checker canary-checker-1.0.0.tgz -n canary-checker --create-namespace'
 
+      - name: Debug
```
moshloop, line 50: Is this still required

yashmehrotra: My bad, removed it

PR #1914, opened 2024-06-19
fix: kubernetes resource fixes

adityathebe:
- [x] only sync canary jobs from the reconciler if the spec has changed otherwise both the reconciler & SyncCanaryJobs tries to schedule the canary job. Sometimes they both schedule it and run the canary twice back-to-back.
- [x] in kubernetes resource checke: prevent the static resources & resources leaking into context after templating.


checks/kubernetes_resource.go
```diff
@@ -50,7 +50,7 @@ func (c *KubernetesResourceChecker) Type() string {
 
 func (c *KubernetesResourceChecker) Run(ctx *context.Context) pkg.Results {
 	var results pkg.Results
-	for _, conf := range ctx.Canary.Spec.KubernetesResource {
+	for _, conf := range ctx.Canary.DeepCopy().Spec.KubernetesResource {
```
adityathebe, line 53: @moshloop I think we should probably make `context.Canary` a private var and always return DeepCopy with a `Canary()` public method.

pkg/jobs/canary/sync.go
```diff
@@ -62,6 +62,7 @@ func findJob(dbCanary pkg.Canary) *job.Job {
 
 func SyncCanaryJob(ctx context.Context, dbCanary pkg.Canary) error {
 	id := dbCanary.ID.String()
+	logger.Debugf("SyncCanaryJob (id=%s name=%s)", dbCanary.ID, dbCanary.Name)
```
moshloop, line 65:
```suggestion
	ctx.Logger.V(2).Infof("SyncCanaryJob (id=%s name=%s)", dbCanary.ID, dbCanary.Name)
```

pkg/jobs/canary/sync.go
```diff
@@ -113,8 +114,10 @@ func SyncCanaryJob(ctx context.Context, dbCanary pkg.Canary) error {
 		ctx.Debugf("Rescheduling %s canary with updated specs", canary)
 		Unschedule(id)
 		newCanaryJob(canaryJob)
+		return nil
 	}
 
+	logger.Debugf("%s was not rescheduled", canary.Name)
```
moshloop, line 120:
```suggestion
	ctx.Logger.V(2).Infof("%s was not rescheduled", canary.Name)
```

PR #1916, opened 2024-06-19
feat: add option to push topology

yashmehrotra:


pkg/api/topology.go
```diff
@@ -33,3 +40,52 @@ func Topology(c echo.Context) error {
 
 	return c.JSON(http.StatusOK, results)
 }
+
+func PushTopology(c echo.Context) error {
```
moshloop, line 44: I think this should go into mission-control directly,

yashmehrotra: @moshloop Should I import canary-checkers db.Persist Functions there then ?

PR #1925, opened 2024-06-21
feat: support selectors in groupby

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/1924


moshloop:
@adityathebe can we add a test for this ?

PR #1929, opened 2024-06-24
add resource selector to group by topology

adityathebe:


moshloop:
@adityathebe can you check the failing tests ?

PR #1965, opened 2024-07-08
feat: push topology after run

yashmehrotra:


api/v1/system_types.go
```diff
@@ -51,6 +56,9 @@ type TopologySpec struct {
 	// Specify the catalog tag (& optionally the tag selector) to group
 	// the topology.
 	GroupBy TopologyTagSelector `json:"groupBy,omitempty"`
+
+	// Agent will push topology to specified path
+	PushLocation PushLocation `json:"pushLocation,omitempty"`
```
moshloop, line 61:
```suggestion
	PushLocation PushLocation `json:"push,omitempty"`
```

pkg/topology/run.go
```diff
@@ -537,6 +540,50 @@ func (tj *TopologyJob) Run(job job.JobRuntime) error {
...
+				job.History.AddErrorf("error pushing topology: %v", err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func pushTopologyToLocation(ctx dutyCtx.Context, target v1.PushLocation, comp pkg.Component) error {
+	ep := fmt.Sprintf("%s/push/topology", target.URL)
```
moshloop, line 555: I don't think we should hardcode the path - You may want to use this with a generic webhook

pkg/topology/run.go
```diff
@@ -537,6 +540,50 @@ func (tj *TopologyJob) Run(job job.JobRuntime) error {
...
+			return fmt.Errorf("error getting password for topology[%s] push: %w", comp, err)
+		}
+
+		client = client.Auth(user, password)
+	}
+
+	req := client.R(ctx).
+		Header(echo.HeaderContentType, echo.MIMEApplicationJSON)
+
+	if canary.UpstreamConf.Valid() {
```
moshloop, line 574: Not sure that these should be linked - we can add the agent name to the url in the chart

api/v1/system_types.go
```diff
@@ -29,6 +29,11 @@ func (t *TopologyTagSelector) IsEmpty() bool {
 	return t.Tag == "" && t.Selector.IsEmpty()
 }
 
+type PushLocation struct {
+	URL  string               `json:"url,omitempty"`
+	Auth types.Authentication `json:"auth,omitempty"`
```
moshloop, line 34:
```suggestion
     types.Authentication `json:"inline,omitempty"`
```

pkg/topology/run.go
```diff
@@ -537,6 +539,44 @@ func (tj *TopologyJob) Run(job job.JobRuntime) error {
...
+				job.History.AddErrorf("error pushing topology: %v", err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func pushTopologyToLocation(ctx dutyCtx.Context, target v1.PushLocation, comp pkg.Component) error {
+	client := http.NewClient()
```
moshloop, line 554: I don't think this caters for oauth / bearer / or tls settings - we should probably extract checks/http.go:generateHTTPRequest into createHttpClient(types.Authentication) or types.Authentication.createCHttpClient()

checks/http.go
```diff
@@ -63,31 +66,54 @@ func (c *HTTPChecker) Run(ctx *context.Context) pkg.Results {
...
 		}
-
-		client.Header(header.Name, value)
+		password, err := ctx.GetEnvValueFromCache(auth.Password, ctx.GetNamespace())
+		if err != nil {
+			return nil, err
+		}
+		client.Auth(user, password)
+	} else if !auth.OAuth.IsEmpty() {
+		client.OAuth(middlewares.OauthConfig{
```
moshloop, line 83: I don't think this will work with client id / secret as env var - I don't think we need bearer / oauth for push connection, but if it is in there it should work. 

We should probably have a `duty/connections/CreateHttpClient(ctx, types.HttpConnection)`  that reuses code from: https://github.com/flanksource/mission-control/blob/main/connection/check.go  throw an error if the connection type doesn't support http clients)


```go
type HttpConnection struct {
      Authentication `yaml:",inline"`
       URL url `yaml:"url,omitempty"
      Bearer   EnvVar `yaml:"bearer,omitempty" json:"bearer,omitempty"`
      OAuth    OAuth  `yaml:"oauth,omitempty" json:"oauth,omitempty"`
      TLS TLSConfig  `yaml:"tls,omitempty" json:"tls,omitempty"`

func (http HttpConnection) Hydrate(ctx Context, namespace) (HttpConnection, error)

models.NewHttpConnection(ctx Context, models.Connection)  (HttpConnection, error)

models.NewConnectionFromHTTP(http HttpConnection)
```

`types.Authentication` probably should not have oauth / bearer as its not supported everywhere. 

We will probably need a similar model for AWS - if we want to add support for profiles / assumed roles / session names / web identities / etc.. - but that can definitely wait

PR #1984, opened 2024-07-11
chore: use envvar for http connection URL

yashmehrotra:


moshloop:
@yashmehrotra this is a breaking change and would require a new apiVersion with crd migration from old to new

PR #2035, opened 2024-08-12
Adding deployment annotations

nalshamaajc:
### Add Pod-Level Annotations to Deployment Object
This PR introduces the ability to add pod-level annotations to the deployment object. By leveraging the **podAnnotations** value, users can now specify custom annotations that will be applied to the pods created by the deployment. This enhancement provides greater flexibility for configuring pod metadata.


chart/values.yaml
```diff
@@ -182,6 +182,10 @@ disableChecks: {}
 # a list of properties to update on startup
 properties: {}
 
+# deployment pod-level annotations
+# podAnnotations: 
+#   some: annotation
+
```
moshloop, line 188:
```suggestion
# @schema
# required: false
# additionalProperties: true
# type: object
# @schema
podAnnotations: {}
```

nalshamaajc: @moshloop Updated

nalshamaajc:
of course! Thanks for the prompt response, @moshloop when will this change be released in a chart version?

Issue #2041, opened 2024-08-13
Canary selection improvements


moshloop:
Add helm chart arguments for:

- canaryNamespaceSelector 
- canaryLabelSelector

This is in addition to `watchNamespace`

Issue #2036, opened 2024-08-12
Setting serviceAccount.rbac.clusterRole: false generates the errors


nalshamaajc:
Setting `serviceAccount.rbac.clusterRole: false` generates the below errors
Version: `1.0.259`
```
W0812 18:38:07.585472 1 reflector.go:535] pkg/mod/k8s.io/client-go@v0.28.8/tools/cache/reflector.go:229: failed to list *v1.Canary: canaries.canaries.flanksource.com is forbidden: User "system:serviceaccount:canary-checker:canary-checker-sa" cannot list resource "canaries" in API group "canaries.flanksource.com" at the cluster scope
E0812 18:38:07.585517 1 reflector.go:147] pkg/mod/k8s.io/client-go@v0.28.8/tools/cache/reflector.go:229: Failed to watch *v1.Canary: failed to list *v1.Canary: canaries.canaries.flanksource.com is forbidden: User "system:serviceaccount:canary-checker:canary-checker-sa" cannot list resource "canaries" in API group "canaries.flanksource.com" at the cluster scope
W0812 18:38:07.585605 1 reflector.go:535] pkg/mod/k8s.io/client-go@v0.28.8/tools/cache/reflector.go:229: failed to list *v1.Topology: topologies.canaries.flanksource.com is forbidden: User "system:serviceaccount:canary-checker:canary-checker-sa" cannot list resource "topologies" in API group "canaries.flanksource.com" at the cluster scope
E0812 18:38:07.585643 1 reflector.go:147] pkg/mod/k8s.io/client-go@v0.28.8/tools/cache/reflector.go:229: Failed to watch *v1.Topology: failed to list *v1.Topology: topologies.canaries.flanksource.com is forbidden: User "system:serviceaccount:canary-checker:canary-checker-sa" cannot list resource "topologies" in API group "canaries.flanksource.com" at the cluster scope
W0812 18:38:08.580997 1 reflector.go:535] pkg/mod/k8s.io/client-go@v0.28.8/tools/cache/reflector.go:229: failed to list *v1.Canary: canaries.canaries.flanksource.com is forbidden: User "system:serviceaccount:canary-checker:canary-checker-sa" cannot list resource "canaries" in API group "canaries.flanksource.com" at the cluster scope
[...]

PR #2045, opened 2024-08-14
feat: namespaced operator

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/2036

resolves: https://github.com/flanksource/canary-checker/issues/2041


chart/values.yaml
```diff
@@ -44,6 +44,17 @@ jsonLogs: true
 # restrict canary-checker to monitor single namespace for canaries.  Leave blank to monitor all namespaces
 canaryNamespace: ""
 
+# Only canaries matching these filters will run.
+# If left empty, all canaries will run.
+canaryFilter:
```
moshloop, line 49: can we use canaryNamespaceSelector and canarySelector to be more consistent with prometheus

chart/templates/rbac.yaml
```diff
@@ -145,3 +146,17 @@ rules:
       - get
       - update
       - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
moshloop, line 150: Why do we need this ?

adityathebe: don't we need the binding?

This is coming from `chart/templates/clusterrolebinding.yaml` which got deleted and merged here.

nalshamaajc:
Thanks for the prompt response, @moshloop when will these changes be released in a chart version?

adityathebe:
@nalshamaajc this should be available on the latest release.

nalshamaajc:
Thank you @adityathebe, when will the next stable release be (v1.0.260)?

adityathebe:
> Thank you @adityathebe, when will the next stable release be (v1.0.260)?

@nalshamaajc we have automatic beta releases that you can test.

And we're a week or 2 away from `v1.0.261`

Issue #2086, opened 2024-08-23
Kubernetes Resource - Display Messages Not Working


dmorrowjc:
The display formatting for Kubernetes Resource checks does not appear to be working. 

If I setup a basic Kubernetes check and format the display, it seems to work fine:
```
apiVersion: canaries.flanksource.com/v1
kind: Canary
metadata:
  name: kube-system-checks
spec:
  interval: 30
  kubernetes:
    - namespace: kube-system
      name: kube-system
      kind: Pod
      resource:
        labelSelector: k8s-app=kube-dns
      namespaceSelector:
        name: kube-system
      display:
        expr: |
          dyn(results).all(x, k8s.isHealthy(x))
```

However, if I use a check like this within the Kubernetes Resource check, nothing is displayed:
```
apiVersion: canaries.flanksource.com/v1
kind: Canary
metadata:
  name: daemonset-checks
  namespace: canary-checker
spec:
  schedule: "@every 1m"
  kubernetesResource:
    - name: daemonset
      namespace: canary-checker
      description: "Create a daemonset and check that it's healthy"
      waitFor:
        expr: 'dyn(resources).all(r, k8s.isHealthy(r))'
        interval: 60s
        timeout: 5m

      resources:
        - apiVersion: apps/v1
          kind: DaemonSet
          metadata:
            name: canary-daemonset
            namespace: canary-checker
            labels:
              app: canary-daemonset
          spec:
            selector:
              matchLabels:
                app: canary-daemonset
            template:
              metadata:
                labels:
                  app: canary-daemonset
              spec:
                containers:
                - name: pause
                  image: "gcr.io/google-containers/pause:3.2" 

      checks:
        - kubernetes:
            - name: canary-daemonset
              namespace: canary-checker

              kind: DaemonSet
              healthy: true
              namespaceSelector:
                name: canary-checker
              display:
                expr: |
                  dyn(results).all(x, k8s.isHealthy(x))
              test:
                expr: dyn(results).all(x, k8s.isHealthy(x))
```
![image](https://github.com/user-attachments/assets/5b384a18-da2e-4445-a2c7-3e90a7585bd0)

If I instead try to configure the display message for the Kubernetes Resource check at the scope of the Canary, I am able to do so...
```
apiVersion: canaries.flanksource.com/v1
kind: Canary
metadata:
  name: daemonset-checks
  namespace: canary-checker
spec:
  schedule: "@every 1m"
  kubernetesResource:
    - name: daemonset
[...]

PR #2093, opened 2024-08-28
feat: add display for kubernetes resource check

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/2086

---

On the display template of a **kubernetes resource check**, a new `display: map[string]string` variable is made available. The `display` variable contains the display output of all of the checks defined within the kubernetes resource check.

A default display expression is also added for **kubernetes resource check** _(which I think should probably be added to all the checks)_.

```
display.keys().map(k, k + ': ' + display[k]).join(' | ')
```

This expression simply concats `<check name>: <display> | <check name>: <display> ...`
Example: here's the display of 3 checks 

image

### DIFF
--- a/api/v1/checks.go
+++ b/api/v1/checks.go
@@ -972,6 +972,16 @@ type KubernetesResourceCheck struct {
 	WaitFor KubernetesResourceCheckWaitFor `json:"waitFor,omitempty"`
 }
 
+func (c KubernetesResourceCheck) GetDisplayTemplate() Template {
+	if !c.Templatable.Display.IsEmpty() {
+		return c.Templatable.Display
+	}
+
+	return Template{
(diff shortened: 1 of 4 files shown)


api/v1/checks.go
```diff
@@ -972,6 +972,16 @@ type KubernetesResourceCheck struct {
 	WaitFor KubernetesResourceCheckWaitFor `json:"waitFor,omitempty"`
 }
 
+func (c KubernetesResourceCheck) GetDisplayTemplate() Template {
+	if !c.Templatable.Display.IsEmpty() {
+		return c.Templatable.Display
+	}
+
+	return Template{
+		Expression: "display.keys().map(k, k + ': ' + display[k]).join(' | ')",
```
moshloop, line 981:
```suggestion
		Expression: "display.keys().map(k, k + ': ' + display[k]).join('\n')",
```

Issue #2059, opened 2024-08-20
Truncate check duration to 2147483647


moshloop:
error adding check to cache: error persisting check status with canary 222c78c0-b593-4e61-b298-c7371ccc6c4e: error adding check status to postgres: failed to encode args[2]: unable to encode 9367476870 into binary format for int4 (OID 23): 9367476870 is greater than maximum value for int4

PR #2181, opened 2024-09-27
feat: limit check status duration to maxInt32 milliseconds.

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/2059


moshloop:
@adityathebe I don't think we want to increase the column size as a very long duration in multiple months is an edge case, and better suited to an analysis then a check

Issue #2246, opened 2024-10-15
Scale down of canaries does not work


moshloop:
Canaries keep running even with replicas: 0

PR #2247, opened 2024-10-16
fix: canary scaling

adityathebe:
resolve: https://github.com/flanksource/canary-checker/issues/2246


pkg/controllers/canary_controller.go
```diff
@@ -142,13 +142,15 @@ func (r *CanaryReconciler) Reconcile(parentCtx gocontext.Context, req ctrl.Reque
 	if canaryForStatus.Status.Replicas != canary.Spec.Replicas {
 		if canary.Spec.Replicas == 0 {
 			canaryJobs.Unschedule(canary.GetPersistedID())
-			if err := db.SuspendCanary(ctx, canary.GetPersistedID(), true); err != nil {
+
+			if err := r.suspendCanary(ctx, canary, true); err != nil {
```
moshloop, line 146: I don't think we should update the suspend annotation based on spec.replicas. If a manual suspend annotation is applied and then the canary is scaled to 0, and then back up again that manually applied annotation will be lost -  spec.replicas=0 should be equivalent to schedule: @never

adityathebe: The problem with relying on `spec.Replicas` is that by default it's always `0`.

Issue #2227, opened 2024-10-11
check.labels should be added to prometheus_metrics


moshloop:
```
  http:
    - url: https://httpbin.demo.aws.flanksource.com/status/200
      name: http-pass-single
      labels:
        check1: http-200
```

`canary_check_info{canary_name="http-checks",canary_namespace="test2",endpoint="https://httpbin.demo.aws.flanksource.com/status/200",key="01927a78-f161-2f8d-252a-1f2d8aa2e159",name="http-pass-single",owner="",severity="",type="http", check1="http-200"} 0`

PR #2257, opened 2024-10-16
feat: use check label in metrics

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/2227


moshloop:
@adityathebe - can we update the chart ?

PR #2261, opened 2024-10-18
feat: canary_check_invalid_count metric

adityathebe:


checks/http.go
```diff
@@ -174,7 +174,7 @@ func (c *HTTPChecker) Check(ctx *context.Context, extConfig external.Check) pkg.
 
 	connection, connectionData, err := ctx.GetConnectionTemplate(check.Connection)
 	if err != nil {
-		return results.Invalidf("error getting connection  %v", err)
+		return results.Failf("error getting connection  %v", err)
```
moshloop, line 177: This should be invalid ?

adityathebe: hmm I guess could be both.

You could run into db timeout or kubernetes API call timeout issue.

PR #2263, opened 2024-10-21
feat: leader election

adityathebe:


cmd/operator.go
```diff
@@ -96,6 +114,7 @@ func run(cmd *cobra.Command, args []string) {
 		Scheme:                  scheme,
 		LeaderElection:          enableLeaderElection,
 		LeaderElectionNamespace: runner.WatchNamespace,
+		LeaderElectionID:        "fa62cd4d.flanksource.com",
```
adityathebe, line 117: The election ID was just copied over from config-db and the first letter was changed.

Issue #2221, opened 2024-10-10
Test check cancellation /  db failure during shutdown


moshloop:
A failed connection to the internal DB (e.g. during DB shutdown) could result in an increase of failed_count

- [ ] Ensure failed db calls result in an invalid and not failed metric 
- [ ] Wait until all running checks have completed (or have been cancelled) before initiating the shutdown of the DB
   - [ ] For checks that require a non-trivial implementation to support cancellation, log backlogged bugs

adityathebe:
> Ensure failed db calls result in an invalid and not failed metric

@moshloop Is that invalid though? I think we should consider invalid if the spec is incorrect.

I think we need to add the distinction between an error and a failure instead.

PR #2290, opened 2024-10-24
feat: check db error

adityathebe:
related: https://github.com/flanksource/canary-checker/issues/2221

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -94,7 +94,15 @@ func setupMetrics() {
 		checkLabels,
 	)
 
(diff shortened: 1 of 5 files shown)


pkg/metrics/metrics.go
```diff
@@ -115,6 +115,7 @@ var (
 	OpsCount        *prometheus.CounterVec
 	OpsFailedCount  *prometheus.CounterVec
 	OpsSuccessCount *prometheus.CounterVec
+	OpsErrorCount   *prometheus.CounterVec
```
moshloop, line 118: Where is this initialized and registered ?

adityathebe: ah looks like it got removed during rebase conflict handling.

PR #2317, opened 2024-11-05
fix(kubernetes_resource): do not copy annotations, last-applied-resource being templated

yashmehrotra:


checks/kubernetes_resource.go
```diff
@@ -113,8 +119,11 @@ func (c *KubernetesResourceChecker) Check(ctx context.Context, check v1.Kubernet
 	displayPerCheck := map[string]string{}
 	for _, c := range check.Checks {
 		virtualCanary := v1.Canary{
-			ObjectMeta: ctx.Canary.ObjectMeta,
-			Spec:       c.CanarySpec,
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      ctx.Canary.ObjectMeta.Name,
+				Namespace: ctx.Canary.ObjectMeta.Namespace,
+			},
```
moshloop, line 125: I think at the very least we should carry through the labels -

yashmehrotra: 👍🏽

Issue #2341, opened 2024-11-21
kubernetes resource check: owner ref invalid


adityathebe:
```
// OwnerReference contains enough information to let you identify an owning
// object. An owning object must be in the same namespace as the dependent, or
// be cluster-scoped, so there is no namespace field.
// +structType=atomic
type OwnerReference struct {
```

we have canaries that can create resource in any namespace. however setting the owner ref isn't possible.

```
Events:
  Type     Reason                    Age   From                          Message
  ----     ------                    ----  ----                          -------
  Normal   Scheduled                 18s   default-scheduler             Successfully assigned default/httpbin-pod-1 to esr
  Warning  OwnerRefInvalidNamespace  19s   garbage-collector-controller  ownerRef [canaries.flanksource.com/v1/Canary, namespace: default, name: pod-svc-test, uid: f24b24ed-e3b0-47e5-be2e-0ef750d9d97a] does not exist in namespace "default"
  Normal   Pulling                   18s   kubelet                       Pulling image "kennethreitz/httpbin:latest"
  Normal   Pulled                    16s   kubelet                       Successfully pulled image "kennethreitz/httpbin:latest" in 2.137582012s (2.137588867s including waiting)
  Normal   Created                   16s   kubelet                       Created container httpbin
  Normal   Started                   16s   kubelet                       Started container httpbin
  Normal   Killing                   16s   kubelet                       Stopping container httpbin
```

PR #2343, opened 2024-11-21
fix: only set the canary in the ownerReferences if the namespace is known & matches the resources's namespace

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/2341

### DIFF
--- a/api/v1/checks.go
+++ b/api/v1/checks.go
@@ -16,6 +16,7 @@ import (
 	"github.com/samber/lo"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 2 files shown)


api/v1/checks.go
```diff
@@ -978,12 +978,40 @@ type KubernetesResourceCheck struct {
 	WaitFor KubernetesResourceCheckWaitFor `json:"waitFor,omitempty"`
 }
 
-func (c *KubernetesResourceCheck) SetCanaryOwnerReference(id, name string) {
+// SetMissingNamespace will set the parent canaries name to resources whose namespace
+// is not explicitly specified.
+func (c *KubernetesResourceCheck) SetMissingNamespace(parent Canary) {
```
adityathebe, line 983: doesn't work with cluster scoped resources

Issue #2373, opened 2024-12-17
Invalid status is never set on canary status.


adityathebe:
```yaml
apiVersion: canaries.flanksource.com/v1
kind: Canary
metadata:
  creationTimestamp: "2024-12-17T06:34:05Z"
  finalizers:
    - canary.canaries.flanksource.com
  generation: 2
  name: invalid-http-check
  namespace: mc
  resourceVersion: "117508601"
  uid: 7f01f959-5adb-48fd-afc5-69ec44e920c7
spec:
  http:
    - display: {}
      endpoint: httplocalhost # <----------------- bad endpoint
      name: bad http endpoint
      password: {}
      responseCodes:
        - 200
      test: {}
      transform: {}
      username: {}
  replicas: 1
  schedule: '@every 24h'
status:
  checkStatus:
    0193d353-1fcc-b669-affd-0838dc369fc2:
      uptime1h: 0/8 0%
  checks:
    bad http endpoint: 0193d353-1fcc-b669-affd-0838dc369fc2
  lastCheck: "2024-12-17T06:36:57Z"
  latency1h: 0ms
  observedGeneration: 2
  replicas: 1
  status: Failed # <----------------- should be Invalid
  uptime1h: 0/8 0%
```

PR #2374, opened 2024-12-17
feat: set "Invalid" status on canary CRD status

adityathebe:
resolves: https://github.com/flanksource/canary-checker/issues/2373


pkg/controllers/canary_controller.go
```diff
@@ -254,11 +254,7 @@ func (r *CanaryReconciler) Report() {
 
 		canary.Status.LastCheck = &metav1.Time{Time: time.Now()}
 		canary.Status.ChecksStatus = payload.CheckStatus
-		if payload.Pass {
-			canary.Status.Status = &v1.Passed
-		} else {
-			canary.Status.Status = &v1.Failed
-		}
+		canary.Status.Status = &payload.Status
```
moshloop, line 257: Can we also check why the message / error message isn't being shown ?

PR #2385, opened 2024-12-24
chore: bump duty to v1.0.800

flankbot:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


moshloop:
@adityathebe can you look at this? I know I said previously to use plurals, but it isn't consistent with the other fields, and reads poorly, this would be better:

```
health: !healthy
status: Progressing,Running,!Terminating
```

PR #2429, opened 2025-02-04
fix: percentile calculation

yashmehrotra:


pkg/metrics/metrics.go
```diff
@@ -293,7 +293,7 @@ func Record(
 
 	_uptime = types.Uptime{Passed: int(pass.Reduce(rolling.Sum)), Failed: int(fail.Reduce(rolling.Sum))}
 	if latency != nil {
-		_latency = types.Latency{Rolling1H: latency.Reduce(rolling.Percentile(95))}
+		_latency = types.Latency{Rolling1H: latency.Reduce(rolling.FastPercentile(99))}
```
moshloop, line 296: Why do we change this to 99?

yashmehrotra: edge case

moshloop: I think we want to handle the edge case e.g.  if 95 = 0 and 99 > 0; 95 = 99

yashmehrotra: Updated

PR #2433, opened 2025-02-07
fix(sql): use string for unkwown types

yashmehrotra:


yashmehrotra:
@moshloop We can ignore the failing test, its due to: https://github.com/microsoft/mssql-docker/issues/895

PR #2440, opened 2025-02-11
feat(s3): allow specifying storage class for PutObject

syhpoon:
Allow configuring S3 check to use a non-default storage class for PutObject.
The storage class is specified as a raw string to allow for new values in the future as well as non-AWS custom storage classes.


adityathebe:
@syhpoon Can you also run `make resources` to generate the CRDs

syhpoon:
> @syhpoon Can you also run `make resources` to generate the CRDs

@adityathebe done!

Issue #2445, opened 2025-02-13
Custom metrics in http check don't work in kubernetesResource


bjan-sag:
Hi!

Now being able to test kubernetesResource, I ran into a different issue :) When trying to use kubernetesResource + http checks, it works fine. But when I try to extract custom metrics from the http check, they don't appear in metrics at all.

Example config (stitched together from https://github.com/flanksource/canary-checker/blob/master/fixtures/k8s/kubernetes_resource_service_pass.yaml and https://github.com/flanksource/canary-checker/blob/master/fixtures/minimal/metrics.yaml):

<details>

<summary>canary.yaml</summary>

```yaml
---
apiVersion: canaries.flanksource.com/v1
kind: Canary
metadata:
  name: pod-svc-test
  namespace: default
  labels:
    "Expected-Fail": "false"
spec:
  schedule: "@every 2m"
  kubernetesResource:
    - name: service accessibility test
      namespace: default
      description: "deploy httpbin & check that it's accessible via its service"
      waitFor:
        expr: 'dyn(resources).all(r, k8s.isReady(r))'
        interval: 2s
        timeout: 2m
      resources:
        - apiVersion: v1
          kind: Pod
          metadata:
            name: httpbin-pod-1
            namespace: default
            labels:
              app: httpbin-pod-1
          spec:
            containers:
              - name: httpbin
                image: "kennethreitz/httpbin:latest"
                ports:
                  - containerPort: 80
        - apiVersion: v1
          kind: Service
          metadata:
            name: httpbin-svc
            namespace: default
          spec:
            selector:
              app: httpbin-pod-1
            ports:
              - port: 80
                targetPort: 80
      checks:
        - http:
            - name: Call httpbin service
              url: "http://httpbin-svc.default.svc"
              metrics:
                - name: httpbin_count
                  type: counter
                  value: "1"
                  labels:
                    - name: check_name
                      valueExpr: check.name
                    - name: code
                      valueExpr: code
                - name: httpbin_2xx_duration
                  type: counter
                  value: elapsed.getMilliseconds()
                  labels:
                    - name: check_name
                      valueExpr: check.name
      checkRetries:
        delay: 2s
        interval: 3s
        timeout: 2m
```

</details>

It shows green in the UI, but when I query metrics the ones I defined are not there:

```
# kubectl -n canary-checker exec canary-checker-c659c5578-mjt5v -- curl -s localhost:8080/metrics | grep httpbin_count
#
```

[...]

PR #2455, opened 2025-02-24
fix: export metrics for checks in kubernetes resources

yashmehrotra:
Fixes: #2445

### DIFF
--- a/checks/runchecks.go
+++ b/checks/runchecks.go
@@ -59,8 +59,8 @@ func getDisabledChecks(ctx *context.Context) (map[string]struct{}, error) {
 	return result, nil
 }
 
-// Exec runs the actions specified and returns the results, without exporting any metrics
-func Exec(ctx *context.Context) ([]*pkg.CheckResult, error) {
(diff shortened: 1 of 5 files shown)


checks/runchecks.go
```diff
@@ -59,8 +59,8 @@ func getDisabledChecks(ctx *context.Context) (map[string]struct{}, error) {
 	return result, nil
 }
 
-// Exec runs the actions specified and returns the results, without exporting any metrics
-func Exec(ctx *context.Context) ([]*pkg.CheckResult, error) {
+// Exec runs the actions specified and returns the results, without saving artifacts
+func Exec(ctx *context.Context, exportMetrics bool) ([]*pkg.CheckResult, error) {
```
yashmehrotra, line 63: This function is used by topology run and kubernetes resource check, was the reason to not export metrics only for topology run to not export metrics if the checks had it defined ?

moshloop: I think in both cases if a custom metric is defined it should be exported - but we shouldn't export the standard check metrics

PR #2543, opened 2025-04-04
feat: Add service monitor configuration with TLS support

mohamedragab2024:
This pull request updates the service monitor resource to allow add TLS configuration to enable HTTPS metrics scrape


chart/templates/servicemonitor.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if eq .Values.serviceMonitor true }}
+{{- if eq .Values.serviceMonitor.enabled true }}
```
moshloop, line 1: Hi @mohamedragab2024 

Thanks for this PR, Can you use `serviceMonitorTlsConfig` - I know that serviceMonitor.enabled is more idiomatic, but it would also be breaking change and them run `make values.schema.deref.json README.md` from the charts folder

mohamedragab2024: Sure fair point, I will update it

mohamedragab2024: @moshloop  I fixed it.

Issue #2588, opened 2025-04-29
Google Cloud Alerts


moshloop:
- Same as alertmanager

PR #2591, opened 2025-05-07
feat: add gcp incidents

yashmehrotra:
Fixes: #2588


pkg/api.go
```diff
@@ -215,7 +215,7 @@ func CheckStatusFromResult(result CheckResult) CheckStatus {
 	cs := CheckStatus{
 		Status:  result.Pass,
 		Invalid: result.Invalid,
-		Time:    time.Now().UTC().Format(time.RFC3339),
+		Time:    time.Now().UTC().Format(time.RFC3339Nano),
```
yashmehrotra, line 218: This was required to return multiple results, previously the accuracy was upto seconds, and due to the unique(check_id, time) clause, only one incident was being added

api/v1/checks.go
```diff
@@ -689,6 +689,32 @@ func (c DNSCheck) GetType() string {
...
+type GCPIncident struct {
+	connection.GCPConnection `yaml:",inline" json:",inline"`
+	Subscription             string `yaml:"subscription" json:"subscription"`
+	Project                  string `yaml:"project" json:"project"`
+}
+
+type PubSubCheck struct {
+	Description  `yaml:",inline" json:",inline"`
+	Templatable  `yaml:",inline" json:",inline"`
+	GCPIncidents GCPIncident `yaml:"gcp_incidents,omitempty" json:"gcp_incidents,omitempty"`
```
moshloop, line 701: I am expecting  just connection objects to the queues here, the pubsub check itself is not aware that its being used for gcp incidents

checks/common.go
```diff
@@ -69,9 +71,20 @@ func template(ctx *context.Context, template v1.Template) (string, error) {
 		tpl.Functions[k] = v
 	}
 
+	tpl.CelEnvs = append(tpl.CelEnvs, context.CelFuncs...)
+
 	return ctx.RunTemplate(tpl, ctx.Environment)
 }
 
+func getDefaultTransformer(check external.Check) v1.Template {
```
moshloop, line 79: this shouldb't be defaulted

checks/gcp_incidents.go
```diff
@@ -0,0 +1,256 @@
...
+
+func CheckGCPIncidents(ctx *context.Context, extConfig external.Check) pkg.Results {
+	var results pkg.Results
+	pubSubCheck := extConfig.(v1.PubSubCheck)
+	check := pubSubCheck.GCPIncidents
+	result := pkg.Success(pubSubCheck, ctx.Canary)
+	results = append(results, result)
+
+	var tokenSrc oauth2.TokenSource
+	if check.ConnectionName != "" {
```
moshloop, line 189: I was expecting this to be a `duty.OpenSubscription(connObject)`

checks/gcp_incidents.go
```diff
@@ -0,0 +1,256 @@
...
+		_ = json.Unmarshal([]byte(rawMsg), &payload)
+
+		inc := payload.Incident
+		// We ignore acknowledgement state
+		switch inc.State {
+		case "closed":
+			// We set the status to true directly in database
+			if ctx.DB() != nil {
+				if pubSubCheck.ResultLookup == "" {
+					pubSubCheck.ResultLookup = `$.incident_id`
```
moshloop, line 242: this should be in the cel function

checks/gcp_incidents.go
```diff
@@ -0,0 +1,256 @@
...
+				}
+				whereClause := fmt.Sprintf(`trim('"' FROM jsonb_path_query_first(details, '%s')::text) = ?`, pubSubCheck.ResultLookup)
+				if err := ctx.DB().Table("check_statuses").Where(whereClause, inc.ID).UpdateColumn("status", true).Error; err != nil {
```
moshloop, line 245: This should filter by transformed=true and canary/check_id

checks/gcp_incidents.go
moshloop: I think it might also make sense todo `stringToUUIDv5(incidentId)`  that we can use then use as a check_id

pkg/api.go
moshloop: We shouldn't need this as each incident should be a different check_id ?

pkg/cache/postgres.go
```diff
@@ -69,10 +69,23 @@ func (c *postgresCache) AddCheckStatus(conn *gorm.DB, check pkg.Check, status pk
 		nextRuntime, _ = check.Canary.NextRuntime(time.Now())
 	}
 
-	if conn.Model(&checks).
-		Clauses(clause.Returning{Columns: []clause.Column{{Name: "id"}}}).
-		Where("canary_id = ? AND type = ? AND name = ?", check.CanaryID, check.Type, check.GetName()).
-		Updates(map[string]any{"status": check.Status, "labels": check.Labels, "last_runtime": status.Time, "next_runtime": nextRuntime}).Error != nil {
+	columnUpdates := map[string]any{
```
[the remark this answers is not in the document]
yashmehrotra: Yeah, thats what we are doing, the only change is that if check_id is present we find the check via id instead of canary/type

Regarding triggers, I didn't want to remove/break existing logic. I can revisit and verify if this query is needed in a different PR or redundant due to triggers

api/v1/common.go
```diff
@@ -400,6 +401,8 @@ type Description struct {
 	// https://canarychecker.io/concepts/metrics-exporter
 	// +kubebuilder:validation:XPreserveUnknownFields
 	Metrics []external.Metrics `json:"metrics,omitempty" yaml:"metrics,omitempty"`
+
+	ResultLookup string `json:"result_lookup,omitempty" yaml:"result_lookup,omitempty"`
```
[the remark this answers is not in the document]
yashmehrotra: No longer required, removed

yashmehrotra:
@moshloop Addressed all the commens

pkg/cache/postgres.go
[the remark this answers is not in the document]
moshloop: With this code the message and details of the update would be lost ?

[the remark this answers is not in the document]
yashmehrotra: Umm, no. This is the checks table, the message and details are stored in the check_statuses table, and we only keep the updated status in checks table

```
+----------------------+--------------------------+----------------------------------------------------------------+
| Column               | Type                     | Modifiers                                                      |
|----------------------+--------------------------+----------------------------------------------------------------|
| id                   | uuid                     |  not null default generate_ulid()                              |
| canary_id            | uuid                     |  not null                                                      |
| agent_id             | uuid                     |  not null default '00000000-0000-0000-0000-000000000000'::uuid |
| type                 | text                     |  not null                                                      |
| name                 | text                     |  not null                                                      |
| namespace            | text                     |                                                                |
| description          | text                     |                                                                |
| icon                 | text                     |                                                                |
| spec                 | jsonb                    |                                                                |
| labels               | jsonb                    |                                                                |
| owner                | text                     |                                                                |
| severity             | text                     |                                                                |
| category             | text                     |                                                                |
| is_pushed            | boolean                  |  not null default false                                        |
| transformed          | boolean                  |                                                                |
| last_runtime         | timestamp with time zone |                                                                |
| last_transition_time | timestamp with time zone |                                                                |
| next_runtime         | timestamp with time zone |                                                                |
| silenced_at          | timestamp with time zone |                                                                |
| status               | text                     |                                                                |
| created_at           | timestamp with time zone |                                                                |
| updated_at           | timestamp with time zone |                                                                |
| deleted_at           | timestamp with time zone |                                                                |
+----------------------+--------------------------+----------------------------------------------------------------+
```

PR #2592, opened 2025-05-07
feat:(sql) Add Timeout to sql check to cancel queries after timeout

mrgb7:
This pull request introduces a new SQL check feature that allows users to define a timeout duration to cancel long-running queries. By default, the timeout is set to 1 minute.


moshloop:
@mrgb7 thanks for the contribution :)

PR #2600, opened 2025-05-26
chore: deprecate httpbin.demo in favour of httpbin.org

yashmehrotra:


moshloop:
I think we had some reliability issues with httpbin.org - Can we not deploy a httpbin in beta / non-prod and redirect the cnmae ?

yashmehrotra:
@moshloop Updated

PR #2802, opened 2025-12-23
feat/check dependencies

mrgb7:
## Motivation

Some use cases require chaining health checks, such as login APIs followed by authenticated requests using tokens from the login response.

## Summary

Adds `dependsOn` field to all check types, enabling checks to run in dependency order and access outputs from previous checks.

## Changes

- Added `dependsOn` field to base `Description` struct (`api/v1/common.go`)
- Added topological sorting in `RunChecks()` for dependency ordering
- Outputs from completed checks available via `outputs.<checkName>.<field>`

## Example

```yaml
http:
  - name: getUuid
    url: https://httpbin.flanksource.com/uuid

  - name: useUuid
    dependsOn:
      - getUuid
    test:
      expr: outputs.getUuid.json.uuid != ""
```

## Validation

- Circular dependency detection
- Missing dependency detection
- Checks with `dependsOn` must have a `name`

## Summary by CodeRabbit

* **New Features**
  * Adds a dependsOn property to many check types and schemas so checks can declare prerequisites.
  * Implements dependency-aware sequencing so checks run in an order that respects declared dependsOn relationships.
  * Exposes per-check outputs so results from earlier checks can be referenced by later ones.

* **Tests**
  * Adds unit tests validating dependency sorting, error cases, and unnamed-check behavior.

* **Chores**
  * Includes an example manifest demonstrating dependsOn usage.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/elastic/go-elasticsearch/v8 v8.19.0
 	github.com/flanksource/artifacts v1.0.18
 	github.com/flanksource/commons v1.43.2
-	github.com/flanksource/duty v1.0.1131
+	github.com/flanksource/duty v1.0.1133
 	github.com/flanksource/gomplate/v3 v3.24.62
 	github.com/flanksource/is-healthy v1.0.82
 	github.com/flanksource/kommons v0.31.7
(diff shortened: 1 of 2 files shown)


moshloop:
Hi @mrgb7 - Thank you for the awesome contribution - this has been requested a couple of times. 

I think this would be useful to all check types so we can extract dependsOn into the base check and order in https://github.com/flanksource/canary-checker/blob/0c9d9f5f8d8688242cfbae25d13973beb8457049/checks/runchecks.go#L92 

And then any check that is depended on, can be saved into https://github.com/flanksource/canary-checker/blob/master/api/context/context.go as outputs{check.name} = [CheckResult](https://github.com/flanksource/canary-checker/blob/master/pkg/api.go#L349)  so that any check that uses context.Template will have access to the check result 

Can you also create fixtures in https://github.com/flanksource/canary-checker/tree/master/fixtures/minimal at least one with a `_pass.yaml` extension as then the fixture testing system will pick it

mrgb7:
@moshloop Can I get +1 review

PR #2844, opened 2026-01-22
chore: make global.serviceMonitor consistent

yashmehrotra:


chart/values.yaml
```diff
@@ -109,7 +109,15 @@ disablePostgrest: false
 # required: false
 # @schema
 # --  Set to true to enable prometheus service monitor
-serviceMonitor: false
+serviceMonitor:
```
moshloop: I think we need to be backwards compatible

PR #3028, opened 2026-07-02
feat(results): classify check failure types

adityathebe:
Introduce a `FailureType` on check results to distinguish why a non-passing check did not pass.

This adds explicit categories for assertion, runtime, evaluation, invalid, and internal failures while keeping legacy `Failf` unclassified so unaudited call sites remain visible during migration.

`ErrorMessage` now classifies runtime failures, expression/template paths classify evaluation failures, `Invalidf` classifies invalid results, and the first classification wins to preserve the primary failure cause.

## Summary by CodeRabbit

* **New Features**
  * Added structured failure categories via a new `failureType` field on check results (assertion, runtime, evaluation/invalid, and internal), including exported failure constants.
  * Added new `Results` failure helper methods to set `failureType`.
* **Bug Fixes**
  * Improved failure classification consistency and ensured the first detected `failureType` takes precedence.
  * Transformer and template-related failures are now categorized more accurately (evaluation vs assertion vs runtime).
  * Metrics for internal-error results now rely on `failureType`.
* **Documentation**
  * Removed the legacy internal-error flag in favor of `failureType == internal`.
* **Tests**
  * Added unit tests covering classification assignment and precedence.
* **Chores**
  * Updated lint installation to use `go install` and added a local binary directory target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -182,7 +182,7 @@ lint: $(GOLANGCI_LINT)
 .PHONY: golangci-lint
 golangci-lint: $(GOLANGCI_LINT)
 $(GOLANGCI_LINT): $(LOCALBIN)
-	test -s $(LOCALBIN)/golangci-lint || curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b $(LOCALBIN) $(GOLANGCI_LINT_VERSION)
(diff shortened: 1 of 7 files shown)


adityathebe:
Scafolding only for easier review. 
Will create another PR that modifies all of the checks.

PR #3032, opened 2026-07-03
fix: replace deprecated Google and Echo APIs

adityathebe:
Fixes #2863.

Staticcheck flagged deprecated Google API credential loading and Echo request logging.

The GCP SQL backup check now uses typed service-account credential JSON, so unexpected credential types are rejected by the Google API client. The Echo server now uses RequestLoggerWithConfig while keeping the existing JSON log fields and avoiding log-write failures replacing request errors.

### DIFF
--- a/checks/database_backup_gcp.go
+++ b/checks/database_backup_gcp.go
@@ -142,7 +142,11 @@ func newSQLAdmin(ctx dutyContext.Context, conn *connection.GCPConnection) (*sqla
 		return nil, err
 	}
 
-	client, err := sqladmin.NewService(ctx.Context, option.WithEndpoint(conn.Endpoint), option.WithCredentialsJSON([]byte(credential))) //nolint:staticcheck
+	client, err := sqladmin.NewService(
+		ctx.Context,
+		option.WithEndpoint(conn.Endpoint),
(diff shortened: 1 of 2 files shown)


checks/database_backup_gcp.go
```diff
@@ -142,7 +142,11 @@ func newSQLAdmin(ctx dutyContext.Context, conn *connection.GCPConnection) (*sqla
 		return nil, err
 	}
 
-	client, err := sqladmin.NewService(ctx.Context, option.WithEndpoint(conn.Endpoint), option.WithCredentialsJSON([]byte(credential))) //nolint:staticcheck
+	client, err := sqladmin.NewService(
+		ctx.Context,
+		option.WithEndpoint(conn.Endpoint),
+		option.WithAuthCredentialsJSON(option.ServiceAccount, []byte(credential)),
```
adityathebe, line 148: Current master accepts 6 credential types, not just service accounts. The deprecated WithCredentialsJSON path flows into cloud.google.com/go/auth's fileCredentials (filetypes.go:43-107 in v0.20.0), which handles service_account, authorized_user (user refresh tokens from gcloud auth application-default login), external_account (workload identity federation), external_account_authorized_user, impersonated_service_account, and gdc_service_account. The deprecated function's own doc comment says "service account or refresh token JSON credentials."

